PR #2, opened 2020-07-21
Add login feature and identify school.

ninaprabhu:


src/main/webapp/script.js
```diff
@@ -0,0 +1,31 @@
...
+      } else {
+      const school = arrStrings[1].slice(start + 1, end);
+      schoolName.innerHTML += 'You go to the following school: ';
+      schoolName.innerHTML +=  school;
+      console.log(school);
+      }
+    }
+    console.log("Displaying login status");
+    loginElement.innerHTML += arrStrings[1];
+
```
lqueipom, line 29: Nit: extra line

src/main/webapp/script.js
```diff
@@ -0,0 +1,31 @@
+function loadBody() {
+  console.log("Verifying login...");
```
lqueipom, line 2: Usually console logs are for testingg, probably should remove them now :)

ninaprabhu: oopsies forgot to delete that, thanks!!

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
-    if (userService.isUserLoggedIn()) {
-      String userEmail = userService.getCurrentUser().getEmail();
-      String urlToRedirectToAfterUserLogsOut = "/login";
-      String logoutUrl = userService.createLogoutURL(urlToRedirectToAfterUserLogsOut);
+    boolean loginStatus = userService.isUserLoggedIn();
+    String urlToRedirect = "/index.html";
 
-      response.getWriter().println("<p>Hello " + userEmail + "!</p>");
-      response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
+    ArrayList<String> arrStrings = new ArrayList();
```
lqueipom, line 36: Alsoo, just saw this but you should change the variable name here :) It took me a bit to figure out what this was supposed to store

ninaprabhu: ooh thanks - totally forgot about that! doing now

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
-      response.getWriter().println("<p>Hello " + userEmail + "!</p>");
-      response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
+    ArrayList<String> userDetails = new ArrayList();
+    if (loginStatus) {
+      userDetails.add("true");
+      String userEmail = userService.getCurrentUser().getEmail();
+      String logoutURL = userService.createLogoutURL(urlToRedirect);
+      userDetails.add(
+          "<p>Hello, "
+              + userEmail
```
hailed748, line 43: If you just added the email and logout URL without the extra html, it could be easier to do the HTML stuff in the JavaScript and would make the parsing a little easier in the JavaScript.

ninaprabhu: Hmm, good point. I wanted to push this as a sign that I have something working, but as I add more stuff to this page, I think I'll definitely make userDetails have more elements and each of them be a separate thing.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
-      String userEmail = userService.getCurrentUser().getEmail();
-      String urlToRedirectToAfterUserLogsOut = "/login";
-      String logoutUrl = userService.createLogoutURL(urlToRedirectToAfterUserLogsOut);
+    boolean loginStatus = userService.isUserLoggedIn();
+    String urlToRedirect = "/index.html";
 
-      response.getWriter().println("<p>Hello " + userEmail + "!</p>");
-      response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
+    ArrayList<String> userDetails = new ArrayList();
+    if (loginStatus) {
```
LukeGMTravis, line 37: `loginStatus` doesn't seem like a very fitting variable name here. Something like `userIsLoggedIn` is better at indicating that this is a boolean for whether or not the user is logged in.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
+      userDetails.add("true");
```
LukeGMTravis, line 38: Adding "true" or "false" is a bit surprising since it seems that this arraylist is for holding html. Perhaps a comment indicating that this is for testing purposes would be appropriate (if that's what it's for).

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
-    if (userService.isUserLoggedIn()) {
-      String userEmail = userService.getCurrentUser().getEmail();
-      String urlToRedirectToAfterUserLogsOut = "/login";
-      String logoutUrl = userService.createLogoutURL(urlToRedirectToAfterUserLogsOut);
+    boolean loginStatus = userService.isUserLoggedIn();
+    String urlToRedirect = "/index.html";
 
-      response.getWriter().println("<p>Hello " + userEmail + "!</p>");
-      response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
+    ArrayList<String> userDetails = new ArrayList();
```
LukeGMTravis, line 36: `userDetails` makes me think that this will have user info in it but it looks like what's actually added below is the html for the response. Perhaps a more descriptive variable name?

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
+    if (loginStatus) {
+      userDetails.add("true");
+      String userEmail = userService.getCurrentUser().getEmail();
+      String logoutURL = userService.createLogoutURL(urlToRedirect);
+      userDetails.add(
```
LukeGMTravis, line 41: Consider using string formatting here: https://dzone.com/articles/java-string-format-examples

LukeGMTravis: And below

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    response.setContentType("text/html");
-
     UserService userService = UserServiceFactory.getUserService();
-    if (userService.isUserLoggedIn()) {
-      String userEmail = userService.getCurrentUser().getEmail();
-      String urlToRedirectToAfterUserLogsOut = "/login";
-      String logoutUrl = userService.createLogoutURL(urlToRedirectToAfterUserLogsOut);
+    boolean loginStatus = userService.isUserLoggedIn();
+    String urlToRedirect = "/index.html";
```
LukeGMTravis, line 34: Nit: `redirectUrl` is a more succinct variable name

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
-      String loginUrl = userService.createLoginURL(urlToRedirectToAfterUserLogsIn);
-
-      response.getWriter().println("<p>Hello stranger.</p>");
-      response.getWriter().println("<p>Login <a href=\"" + loginUrl + "\">here</a>.</p>");
+      userDetails.add("false");
+      String loginURL = userService.createLoginURL(urlToRedirect);
+      userDetails.add("<p><a href=\"" + loginURL + "\"> Login Here </a></p>");
     }
+
+    Gson gson = new Gson();
```
LukeGMTravis, line 54: You can do this part in line down below, ie: `new Gson().toJson(...`

src/main/webapp/index.html
```diff
@@ -6,7 +6,8 @@
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
-  <body>
-      <p> Hello </p>
+  <body onload = "loadBody()">
```
LukeGMTravis, line 9: Nit: spacing

src/main/webapp/script.js
LukeGMTravis, line 2: This is probably okay since it's prototype code but in the future you won't want to console log stuff. Maybe good to leave a comment here reminding you to remove this later.

LukeGMTravis: Ditto below

src/main/java/com/google/sps/servlets/LoginServlet.java
LukeGMTravis: Ah nvm, just checked how you're using this below. This should definitely be sent as a json object, not as an array. It's much more error prone to try to remember which index is which value then to have named keys : )

src/main/webapp/script.js
```diff
@@ -0,0 +1,31 @@
...
+  verifyLogin();
+}
+
+/** Contacts Login servlet to verify if user is logged in. */
+function verifyLogin() {
+  console.log("Fetching login status...");
+  fetch('/login').then(response => response.json()).then((arrStrings) => {
+    const loginElement = document.getElementById("login-box");
+    if (arrStrings[0] == "false") {
+      document.getElementById("school-name").style.display = "none";
```
LukeGMTravis, line 12: This is probably unnecessary if you're just going to redirect the user to a login page

src/main/webapp/script.js
```diff
@@ -0,0 +1,31 @@
...
+    } else {
+      const schoolName = document.getElementById("school-name");
+      schoolName.style.display = "block";
+      const start = arrStrings[1].indexOf('@');
```
LukeGMTravis, line 16: This isn't a great way to figure out if the user's email is valid. This kind of into should be returned from the server specifically, not packaged inside some html.

LukeGMTravis: Ditto for school.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -27,22 +29,30 @@
...
 
-      response.getWriter().println("<p>Hello " + userEmail + "!</p>");
-      response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
+    ArrayList<String> userDetails = new ArrayList();
+    if (loginStatus) {
+      userDetails.add("true");
+      String userEmail = userService.getCurrentUser().getEmail();
+      String logoutURL = userService.createLogoutURL(urlToRedirect);
+      userDetails.add(
+          "<p>Hello, "
```
LukeGMTravis, line 42: Also, generally the server doesn't specify strings like this. After reading your JS below it seems like it would be much more appropriate to just return the userEmail and any other data inside a json object and have the JS construct the string to show to the user.

target/step159-2020-1/index.html
```diff
@@ -6,7 +6,8 @@
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
-  <body>
-      <p> Hello </p>
+  <body onload = "loadBody()">
```
LukeGMTravis, line 9: Spacing

target/step159-2020-1/script.js
```diff
@@ -0,0 +1,27 @@
+function loadBody() {
```
LukeGMTravis, line 1: This looks like duplicate code? Do you mean to have two copies of the same code?

src/main/webapp/script.js
```diff
@@ -0,0 +1,31 @@
...
+      const end = arrStrings[1].indexOf('.');
+      if (start == -1 || end == -1) {
+          schoolName.innerHTML += "Sorry, you have entered an invalid email."
+      } else {
+      const school = arrStrings[1].slice(start + 1, end);
```
LukeGMTravis, line 21: Nit: indentation

src/main/java/com/google/sps/servlets/LoginServlet.java
ninaprabhu: ^ just kidding, decided to implement this because it'll make it easier in future

src/main/java/com/google/sps/servlets/LoginServlet.java
ninaprabhu: makes sense, implemented it as a userDetails property!

src/main/java/com/google/sps/servlets/LoginServlet.java
ninaprabhu: I just added email and school name- hopefully that makes more sense? if not, I can change it to loginDetails. I intend for this to later hold the user's name and other facts that they might put in (year in school, etc)

src/main/webapp/script.js
ninaprabhu: I want to use OAuth for this! However, I can't enable it right now - Travis mentioned he's going to look into it and I'm going to try again now that we have our project ID. I just wanted to have a rudimentary version working so I had some form of error checking (even though it's not great)

target/step159-2020-1/script.js
ninaprabhu: Sorry, where is this duplicated?

src/main/webapp/script.js
ninaprabhu: ^also Laura mentioned the trumail API which I'll look into

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -21,28 +21,38 @@
...
 
-      response.getWriter().println("<p>Hello " + userEmail + "!</p>");
-      response.getWriter().println("<p>Logout <a href=\"" + logoutUrl + "\">here</a>.</p>");
+    JSONObject userDetails = new JSONObject();
+    userDetails.put("loggedIn", loggedIn);
+    if (loggedIn) {
+      String userEmail = userService.getCurrentUser().getEmail();
+      userDetails.put("userEmail", userEmail);
+      String logoutURL = userService.createLogoutURL(urlToRedirect);
+      userDetails.put("url", logoutURL);
```
LukeGMTravis, line 41: Just "url" is pretty non-descriptive of what this is.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -21,28 +21,38 @@
...
+      int start = userEmail.indexOf('@');
+      int end = userEmail.indexOf('.');
```
LukeGMTravis, line 43: Technically this may not work as you expect. Consider the following email: "luke.travis@mit.edu"

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -21,28 +21,38 @@
...
+      String schoolName;
+      if (start == -1 || end == -1) { // This means that @ or . are not in userEmail.
+        schoolName = "Sorry, you have entered an invalid email.";
```
LukeGMTravis, line 46: The better practice here is to just not set the school name

src/main/webapp/script.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/** Contacts Login servlet to verify if user is logged in. */
+function verifyLogin() {
+  fetch('/login').then(response => response.json()).then((userDetails => {
+    const loginElement = document.getElementById("login-box");
+    const schoolName = document.getElementById("school-name");
+    if (!userDetails.loggedIn) {
+      schoolName.style.display = "none";
+      loginElement.innerHTML += "<p><a href=\"" + userDetails.url + "\"> Click here to log in! </a></p>";
```
LukeGMTravis, line 12: Best practice is to use string literals here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/main/webapp/script.js
```diff
@@ -0,0 +1,40 @@
...
+      loginElement.innerHTML += "<p><a href=\"" + userDetails.url + "\"> Click here to log out! </a></p>";
+    }
+  }));
+}
+
+function showClasses() {
+  const classes = document.getElementById("search-results");
+  classes.innerHTML = "";
+  fetch('/search')
+  .then(response => response.json())
```
LukeGMTravis, line 25: Nit: indentation

src/main/webapp/script.js
```diff
@@ -0,0 +1,40 @@
...
+      loginElement.innerHTML += "<p><a href=\"" + userDetails.url + "\"> Click here to log in! </a></p>";
+    } else {
+      schoolName.style.display = "block";
+      schoolName.innerHTML += userDetails.schoolName;
+      loginElement.innerHTML += "<p><a href=\"" + userDetails.url + "\"> Click here to log out! </a></p>";
+    }
+  }));
+}
+
+function showClasses() {
```
LukeGMTravis, line 21: Consider making this a separate PR

src/main/webapp/script.js
```diff
@@ -0,0 +1,40 @@
...
+  const classes = document.getElementById("search-results");
+  classes.innerHTML = "";
+  fetch('/search')
+  .then(response => response.json())
+  .then((response) => {
+    return JSON.parse(response.classNames);
+  })
+  .then((classNames) => {
+    console.log(classNames);
+    classNames.forEach(name => createListElement(name));
```
LukeGMTravis, line 31: This doesn't actually insert the list elements

target/step159-2020-1/script.js
LukeGMTravis: This used to look fairly identical to the code in src/main/webapp/script.js

What's the purpose of the changes in this file?

src/main/webapp/script.js
ninaprabhu: I wanted to because it was more of the search feature - but wasn't sure how to not commit certain changes in a file. Will look into that

src/main/webapp/script.js
ninaprabhu: Realized that - I changed it to append the element to classNames, but I think I committed that in the search branch instead

target/step159-2020-1/script.js
ninaprabhu: I had to push my current code to restructure our filesystem a few days ago - maybe that's what you mean? But since I didn't get that reviewed, I wanted to commit it now to make sure everything looked good with it before formally using it.

src/main/java/com/google/sps/servlets/LoginServlet.java
ninaprabhu: Makes sense, will negate this OR statement and only set school name if the right conditions are met

src/main/java/com/google/sps/servlets/LoginServlet.java
ninaprabhu: This is just kind of a holder until I can get OAuth running - one of my goals for the week

LukeGMTravis: Maybe add in a comment that this should be reworked in the future then. Incidentally that fix here is pretty easy: "luke.travis@mit.edu" -> "mit.edu" -> "mit"

src/main/webapp/script.js
LukeGMTravis: This won't block this time around but it's good practice to have different branches for different features you're working on.

target/step159-2020-1/script.js
LukeGMTravis: So which code path is actually being used? These files in `target/step159-2020-1` or the files above in `src/main/webapp`?

ninaprabhu: I put the target folder in .gitignore, not sure why it's still showing up. I think all of us are having an issue with those files being duplicated in target even though we're working in src/main/webapp

src/main/java/com/google/sps/servlets/LoginServlet.java
ninaprabhu: good point, i'll change the code to reflect that approach

PR #3, opened 2020-07-21
Adds Hardcode Charts

hailed748:
- Adds: This adds hard-coded charts to the index.html page

- Modifies: This modifies index.html, script.js, and adds two additional JavaServlets


src/main/java/com/google/sps/data/EntryObject.java
```diff
@@ -0,0 +1,46 @@
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EntryObject {
```
lqueipom, line 6: Maybe it would be good to have a more descriptive variable in this case (so others know what kind of entry you are referring to)

src/main/java/com/google/sps/data/EntryObject.java
```diff
@@ -0,0 +1,46 @@
...
+  List<Object> hourList = new ArrayList<>();
+  List<Object> difficultyList = new ArrayList<>();
+  List<String> container = new ArrayList();
+
+  public EntryObject(String hourTitle, String diffTitle) {
```
lqueipom, line 11: Saw that you did not use diffTitle, do you need to pass it as a param?

src/main/java/com/google/sps/data/EntryObject.java
```diff
@@ -0,0 +1,46 @@
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class EntryObject {
+  List<Object> hourList = new ArrayList<>();
+  List<Object> difficultyList = new ArrayList<>();
+  List<String> container = new ArrayList();
```
lqueipom, line 9: You initialize container here but also use a different container variable inside the different functions. Would change the name if it's necessary to have it initialized outside.

src/main/java/com/google/sps/data/EntryObject.java
```diff
@@ -0,0 +1,46 @@
...
+    hourList.add(container);
+  }
+
+  public void addHour(String newHour) {
+    List<Integer> container = new ArrayList();
+    container.add(Integer.parseInt(newHour));
+    hourList.add(container);
+  }
+
+  public void addDifficulty(String newDifficulty) {
```
lqueipom, line 33: On the addThing methods, can you not directly add to hourList/difficultyList?

src/main/webapp/script.js
```diff
@@ -0,0 +1,35 @@
+google.charts.load("current", {packages:["corechart"]});
+google.charts.setOnLoadCallback(makeGraph);
+function makeGraph() {
+  fetch('/graph')
+    .then((response) => response.json())
+    .then((data) => {
+      console.log(data);
```
lqueipom, line 7: Delete the console log here if you are done testingg

src/main/webapp/script.js
```diff
@@ -0,0 +1,35 @@
...
+          vAxis: {title: "# Students"},
+          hAxis: {title: "Hours"},
+          histogram: {
+            hideBucketItems: true
+          }
+        };
+        const hourChart = new google.visualization.Histogram(document.getElementById('chart_div'));
+        hourChart.draw(hourData, hourOptions);
+
+
```
lqueipom, line 21: Nit: extra line.

target/step159-2020-1/script.js
```diff
@@ -0,0 +1,8 @@
+function makeGraph() {
+  fetch('/graph')
+    .then((response) => response.json())
+    .then((data) => {
+      dataObject = JSON.parse(data);
+      console.log(dataObject);
```
lqueipom, line 6: Why are there two hours.csv, difficulty.csv, and script.js?

src/main/java/com/google/sps/data/EntryObject.java
```diff
@@ -0,0 +1,46 @@
...
+    container.add(hourTitle);
+    hourList.add(container);
+
+    container.removeAll(container);
+    container.add(hourTitle);
+    difficultyList.add(container);
+  }
+
+  public void addTitle(String title) {
+    List<String> container = new ArrayList();
```
ninaprabhu, line 22: maybe name these titleContainer, hourContainer, etc. just so the number of variables named container doesn't confuse anyone?

src/main/webapp/difficulty.csv
```diff
@@ -0,0 +1,100 @@
+6
```
ninaprabhu, line 1: just something to note - do you expect students to enter difficulty on a sliding scale or freehand in a form? if it's the second, then you might want to add some strings (non-numbers basically) to this .csv and come up with a way to handle them. since you're using parseInt, I think (?) it would crash on anything that wasn't an int.

src/main/webapp/index.html
```diff
@@ -2,11 +2,14 @@
...
+    <title>Capstone</title>
     <link rel="stylesheet" href="style.css">
+    <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
     <script src="script.js"></script>
   </head>
-  <body>
+  <body onload="makeGraph()">
       <p> Hello </p>
+    <div id="chart_div" style="width: 900px; height: 500px;"></div>
+    <div id="chart_div2" style="width: 900px; height: 500px;"></div>
```
ninaprabhu, line 13: if these are for different charts, make the div names more descriptive whenever you decide what they're used for

src/main/java/com/google/sps/data/EntryObject.java
hailed748: I can't because I need to format it in bracket because the chart API needs a nested list format like : [[item1],[item2]....]

target/step159-2020-1/script.js
hailed748: Could you clarify what you mean here. On my end I only one of each file.

src/main/webapp/difficulty.csv
hailed748: csv are actually all string based, that is why I parse them an integers so I don't think it will crash anything.

ninaprabhu: I meant if it's open-ended and someone puts in "hello" or another symbol on accident instead of a number - probably isn't something you need to worry about now though haha

target/step159-2020-1/script.js
lqueipom: Resolved think it was I got confused :)

src/main/java/com/google/sps/data/GraphDataObject.java
```diff
@@ -0,0 +1,39 @@
...
+public class GraphDataObject {
+  List<Object> hourList = new ArrayList<>();
+  List<Object> difficultyList = new ArrayList<>();
+
+  public GraphDataObject(String hourTitle, String diffTitle) {
+    List<String> titleContainer = new ArrayList();
+    titleContainer.add(hourTitle);
+    hourList.add(titleContainer);
+
+    titleContainer.removeAll(titleContainer);
```
LukeGMTravis, line 15: What does this do?

src/main/java/com/google/sps/data/GraphDataObject.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+
+public class GraphDataObject {
+  List<Object> hourList = new ArrayList<>();
+  List<Object> difficultyList = new ArrayList<>();
+
+  public GraphDataObject(String hourTitle, String diffTitle) {
+    List<String> titleContainer = new ArrayList();
+    titleContainer.add(hourTitle);
+    hourList.add(titleContainer);
```
LukeGMTravis, line 13: If you're just needing to turn "hourTitle" into an array to add it to the list then I'm not sure you need to create the "titleContainer" variable to do that, seems you could just do this in place here.

LukeGMTravis: Same comment below for addHour and addDifficulty

src/main/java/com/google/sps/servlets/ChartsServlet.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/graph")
+public class ChartsServlet extends HttpServlet {
+  private List<Object> entryList = new ArrayList<>();
+  GraphDataObject newEntry = new GraphDataObject("Hours", "Difficulty");
+
+  @Override
+  public void init() {
+    Scanner scanner = new Scanner(getServletContext().getResourceAsStream("/hours.csv"));
```
LukeGMTravis, line 21: Shouldn't this be "scannerHours"?

src/main/java/com/google/sps/servlets/ChartsServlet.java
```diff
@@ -0,0 +1,54 @@
...
+      String line = scanner.nextLine();
+      newEntry.addHour(line);
+    }
+    scanner.close();
+
+    while (scannerDifficulty.hasNextLine()) {
+      String line = scannerDifficulty.nextLine();
+      newEntry.addDifficulty(line);
+    }
+    scanner.close();
```
LukeGMTravis, line 35: Wrong scanner

src/main/webapp/difficulty.csv
LukeGMTravis: Good thing to think about. Ideally, I think, your frontend would only allow numbers to be sent (which the server would then verify before storing them). This is good for now though

src/main/webapp/script.js
```diff
@@ -0,0 +1,33 @@
+google.charts.load("current", {packages:["corechart"]});
+google.charts.setOnLoadCallback(makeGraph);
```
LukeGMTravis, line 2: Question: do you need the onload in the html if you have this? Or vice-versa?

src/main/webapp/script.js
```diff
@@ -0,0 +1,33 @@
...
+function makeGraph() {
+  fetch('/graph')
```
LukeGMTravis, line 4: Nit: Consistently use either single or double quotes, same below

src/main/webapp/difficulty.csv
hailed748: Got it. I can handle that in code or we could make the input interface numbers only.

target/step159-2020-1/script.js
LukeGMTravis: I'm seeing the same two hours.csv, difficulty.csv, and script.js

src/main/java/com/google/sps/data/GraphDataObject.java
hailed748: The Charts API says I need to have a name for the data so I need to have it as the first element of the data array. Also the format has to be [[title],[data1],[data2],....] that is why I am wrapping each data point in an array.

hailed748:
No clue why double files are showing up. When I commit it is always one of each.

LukeGMTravis:
> No clue why double files are showing up. When I commit it is always one of each.

It's likely from a prior commit I'd suspect. Just undo the changes in the redundant file.

src/main/java/com/google/sps/data/GraphDataObject.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class GraphDataObject {
+  List<Object> hourList = new ArrayList<>();
+  List<Object> difficultyList = new ArrayList<>();
+
+  public GraphDataObject(String hourTitle, String diffTitle) {
+    hourList.add(Arrays.asList(hourTitle));
```
LukeGMTravis, line 12: So much smaller now : )

hailed748:
Very weird, I had duplicate files in some random folder within a folder called target, not even sure why that is there seems like something maven needs.

PR #4, opened 2020-07-22
Search

ninaprabhu:
NOTE: ONLY LOOK AT THE "IMPLEMENT SEARCH FEATURE" COMMIT TO SEE RELEVANT CHANGED FILES
This version of search lets users upload classes with professor names (single class -> single professor), and find all of the classes associated with a certain professor. I'm planning on having more fields (such as days of the week the class is taught on, number of credit hours, school the course is taught at, etc) and implement functionality for multiple professors per class and other things


src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,80 @@
...
+  /* Show classes. */
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> classNames = new ArrayList<>();
+    List<String> professorNames = new ArrayList<>();
+    String profName = request.getParameter("profName");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Filter propertyFilter = new FilterPredicate("Professor", FilterOperator.EQUAL, profName);
+    Query q = new Query("Class").setFilter(propertyFilter);
+    List<Entity> classes = datastore.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    System.out.println(classes);
```
hailed748, line 49: Nit : Don't forget to remove later, same with the one a couple lines below

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,80 @@
...
+public class SearchServlet extends HttpServlet {
+
+  @Override
+  /* Show classes. */
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> classNames = new ArrayList<>();
+    List<String> professorNames = new ArrayList<>();
+    String profName = request.getParameter("profName");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Filter propertyFilter = new FilterPredicate("Professor", FilterOperator.EQUAL, profName);
```
hailed748, line 46: Just for my understanding how exactly does this work? propertyFilter seems to be a filter of professor entities with the same name as profName. But how can this filter be used in this line: 
   Query q = new Query("Class").setFilter(propertyFilter);

ninaprabhu: Basically it creates a Query looking for things of property Class that also satisfy the filter. Documentation is https://cloud.google.com/datastore/docs/concepts/queries if that's helpful!

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: oops - thanks!

target/step159-2020-1/script.js
```diff
@@ -0,0 +1,42 @@
...
+  .then((classNames) => {
+    classNames.forEach(name => classes.appendChild(createListElement(name)));
+  });
+}
+
+/* Creates an <li> element containing text. */
+function createListElement(text) {
+  const liElement = document.createElement('li');
+  liElement.innerText = text;
+  return liElement;
```
lqueipom, line 41: For some reason it is creating duplicates of the js, html files, etc I implemented a gitignore for this so please let me know if after my PR is merged you still get stuff from the folder targer commited! :)

lqueipom: Actually just saw that you added one as well, great!!

LukeGMTravis:
Use gitignore to avoid committing all these libraries

LukeGMTravis:
I'm going to wait to review this until after your other PR is merged in. As it is now it's kinda hard to tell what you've actually changed in this branch since it has all your commits about login in here as well (in the future, especially if these are separate features, it's best to keep them in completely separate branches and not have them share commits that aren't relevant to both features).

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,115 @@
...
+      Filter profFilter = new FilterPredicate("Professor", FilterOperator.EQUAL, profName);
+      filters.add(profFilter);
+    }
+    List<Integer> units = new ArrayList<>();
+    if (!request.getParameter("units").isEmpty()) {
+      List<String> strUnits = Arrays.asList(request.getParameter("units").split(","));
+      for (String number : strUnits) {
+        units.add(Integer.valueOf(number));
+      }
+      if (units.size() != 0) {
```
LukeGMTravis, line 63: Use "isEmpty"

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,115 @@
...
+      String className = request.getParameter("className");
+      Filter classFilter = new FilterPredicate("Name", FilterOperator.EQUAL, className);
+      filters.add(classFilter);
+    }
+    if (!request.getParameter("profName").isEmpty()) {
+      String profName = request.getParameter("profName");
+      Filter profFilter = new FilterPredicate("Professor", FilterOperator.EQUAL, profName);
+      filters.add(profFilter);
+    }
+    List<Integer> units = new ArrayList<>();
```
LukeGMTravis, line 57: Since this is only used inside the if statement below you should define this inside of it.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,115 @@
...
+      if (units.size() != 0) {
+        Filter unitsFilter = new FilterPredicate("Units", FilterOperator.IN, units);
+        filters.add(unitsFilter);
+      }
+    }
+
+    Query q = new Query("Class");
+    // Add filters depending on number of fields specified
+    if (filters.size() == 1) {
+      q.setFilter(filters.get(0));
```
LukeGMTravis, line 72: Looks like you can just pass in "filters": https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Query.CompositeFilterOperator#and_2

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,115 @@
...
+      for (String number : strUnits) {
+        units.add(Integer.valueOf(number));
+      }
+      if (units.size() != 0) {
+        Filter unitsFilter = new FilterPredicate("Units", FilterOperator.IN, units);
+        filters.add(unitsFilter);
+      }
+    }
+
+    Query q = new Query("Class");
```
LukeGMTravis, line 69: As succinct as just "q" is, I think something more descriptive like "classQuery" might be more helpful (especially if you have different queries in the future).

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,115 @@
...
+      classNames.add(name);
+      String professor = (String) entity.getProperty("Professor");
+      professorNames.add(professor);
+    }
+
+    // Generate links for class pages and send them back - set variables in URL
+    String classNameJson = new Gson().toJson(classNames);
+    String profNameJson = new Gson().toJson(professorNames);
+    JSONObject obj = new JSONObject();
+    obj.put("classNames", classNameJson);
```
LukeGMTravis, line 94: Since the class name and professor name are directly related I think it make more sense to package them together as an object and then return a list of those objects. As it is now you have to form that relationship on the client side by matching indexes, not ideal.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,115 @@
...
+    response.setContentType("application/json;");
+    response.getWriter().println(obj);
+  }
+
+  /* Store class. */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String name = request.getParameter("class-name");
+    String prof = request.getParameter("prof-name");
+    Integer units = Integer.valueOf(request.getParameter("num-units"));
```
LukeGMTravis, line 105: This is okay for now but in the future you'll have to think about how to handle a case where a non-integer is returned.

src/main/webapp/script.js
```diff
@@ -0,0 +1,37 @@
+function countUnits(){
+  const inputElems = document.getElementsByName("search-units");
+  const arr = []
+  for (let i=0; i<inputElems.length; i++) {
```
LukeGMTravis, line 4: Nit: spacing

src/main/webapp/script.js
```diff
@@ -0,0 +1,37 @@
...
+  for (let i=0; i<inputElems.length; i++) {       
+    if (inputElems[i].checked){
+      arr.push(inputElems[i].value);
+    }
+  }
+  return arr;
+}
+
+function showClasses() {
+  const classes = document.getElementById("search-results");
```
LukeGMTravis, line 13: This variable should be more specific considering all the class terminology in this method. Maybe something like "classResults" or "classResultsContainer"

src/main/webapp/script.js
```diff
@@ -0,0 +1,37 @@
...
+  const className = document.getElementById("search-class").value;
+  const profName = document.getElementById("search-prof").value;
+  const units = countUnits();
```
LukeGMTravis, line 16: Nice job splitting up this functionality.

src/main/webapp/script.js
```diff
@@ -0,0 +1,37 @@
...
+  const url = new URL("/search", window.location.origin);
+  url.searchParams.set("className", className);
+  url.searchParams.set("profName", profName);
+  url.searchParams.set("units", units);
+  fetch(url)
+  .then(response => response.json())
+  .then((response) => {
+    return JSON.parse(response.classNames);
+  })
+  .then((classNames) => {
```
LukeGMTravis, line 27: It doesn't look like just a list of class names is returned? Correct me if I'm missing something though

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: For "and", you have to have at least two filters in the list apparently (according to the error I just got), but the documentation led me to believe that "and" took exactly two arguments, so this is actually super helpful. Thanks!

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: Just made this feature a radio button to avoid this issue. I had check boxes for the search feature but just hadn't gotten around to mirroring that in the input feature yet since I wasn't sure if Dagm had done something similar yet. Thanks!

src/main/webapp/script.js
ninaprabhu: Thank you! :)

src/main/webapp/script.js
ninaprabhu: In my servlet, I generate a list of class names and set them to the classNames attribute of my json response

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: Ah makes sense, if I make a class for this now it'll also allow me to add more/change attributes as we start including more details in the search filter

src/main/webapp/script.js
ninaprabhu: There are more things being returned but just as proof of correctness, I am looking at classNames

src/main/java/com/google/sps/servlets/SearchServlet.java
LukeGMTravis: Really, it errors on passing a collection with just a single element? Interesting... Well at least this will help simplify the other cases

src/main/java/com/google/sps/servlets/SearchServlet.java
LukeGMTravis: A class could be a good idea but keep in mind that you'll have to convert that class to a JSON object to send to the client side

src/main/java/com/google/sps/servlets/SearchServlet.java
LukeGMTravis: That's true but people can always query your server directly without using your client : )

src/main/webapp/script.js
LukeGMTravis: Oooh, now I'm seeing that you're only parsing response.classNames, missed that earlier

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: Ooh good point - didn't think about that

ninaprabhu: I'll add in error checking and testing later this week most likely - just wanna get search working and robus for the proper inputs first

#.gitignore#
```diff
@@ -0,0 +1 @@
+target/
```
LukeGMTravis, line 1: Is this file needed?

src/main/java/com/google/sps/servlets/ClassObject.java
```diff
@@ -0,0 +1,19 @@
+package com.google.sps.servlets;
+
+public class ClassObject {
```
LukeGMTravis, line 3: Looking at how this is used you could probably use a constructor here.

LukeGMTravis: Also, class is an ambiguous word, in your design doc you have course/term instead. Please change here and below

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,111 @@
...
+      if (!units.isEmpty()) {
+        Filter unitsFilter = new FilterPredicate("Units", FilterOperator.IN, units);
+        filters.add(unitsFilter);
+      }
+    }
+
+    Query classQuery = new Query("Class");
+    if (!filters.isEmpty()) {
+      if (filters.size() == 1) {
+        classQuery.setFilter(filters.get(0));
```
LukeGMTravis, line 70: Still makes me sad you have to do this :. (

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,111 @@
...
+      } else {
+        classQuery.setFilter(CompositeFilterOperator.and(filters));
+      }
+    }
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    List<Entity> results =
+        datastore.prepare(classQuery).asList(FetchOptions.Builder.withDefaults());
+    for (Entity entity : results) {
+      ClassObject curr = new ClassObject();
```
LukeGMTravis, line 80: "curr" isn't a terribly descriptive variable name

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,111 @@
...
+      String name = (String) entity.getProperty("Name");
+      curr.name = name;
+      String professor = (String) entity.getProperty("Professor");
+      curr.professor = professor;
+      Long numUnits = (Long) entity.getProperty("Units");
+      curr.units = numUnits;
+      classes.add(curr);
+      System.out.println(curr);
```
LukeGMTravis: Left over

src/main/webapp/index.html
```diff
@@ -1,12 +1 @@
...
-    <meta charset="UTF-8">
-    <title>Nina's Portfolio</title>
-    <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
-  </head>
-  <body>
-      <p> Hello </p>
-  </body>
-</html>
+
```
LukeGMTravis, line 1: Do you want to just delete this file instead?

src/main/webapp/script.js
```diff
@@ -0,0 +1,35 @@
...
+  const units = countUnits();
+  classResults.innerHTML = "";
+  const url = new URL("/search", window.location.origin);
+  url.searchParams.set("className", className);
+  url.searchParams.set("profName", profName);
+  url.searchParams.set("units", units);
+  fetch(url)
+  .then(response => response.json())
+  .then((classes) => {
+    console.log(classes);
```
LukeGMTravis, line 25: Left over

src/main/webapp/search.html
```diff
@@ -0,0 +1,42 @@
...
+      <br/>
+      <textarea name="prof-name">Professor Name</textarea>
+      <br/>
+      <fieldset>      
+      <legend>How many units?</legend>      
+      <input type="radio" name="num-units" value="1">1<br>      
+      <input type="radio" name="num-units" value="2">2<br>      
+      <input type="radio" name="num-units" value="3">3<br>      
+      </fieldset> 
+      <br/><br/>
```
LukeGMTravis, line 26: Multiple br tags in a row is an indication that you should start using css to style margin/padding. Same below

#.gitignore#
ninaprabhu: Not even sure how #.gitignore# appeared, will delete

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: agreed :(

src/main/java/com/google/sps/servlets/ClassObject.java
ninaprabhu: I changed every reference to Course to make it more easily readable :)

src/main/webapp/index.html
ninaprabhu: I think if I delete index.html,  I get an error when running my dev server because it doesn't know where to go upon startup. I usually just have a blank file for this and then navigate to search.html

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,118 @@
...
+      String term = request.getParameter("term");
+      Filter termFilter = new FilterPredicate("Term", FilterOperator.EQUAL, term);
+      filters.add(termFilter);
+    }
+
+    List<Integer> units = new ArrayList<>();
+    if (!request.getParameter("units").isEmpty()) {
+      List<String> strUnits = Arrays.asList(request.getParameter("units").split(","));
+      for (String number : strUnits) {
+        units.add(Integer.valueOf(number));
```
hailed748, line 68: You use longs below and because datastore stores numbers as longs maybe use longs here too

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: Yeah - I think I got that up and running, it ended up working out pretty simply

src/main/java/com/google/sps/servlets/Course.java
```diff
@@ -0,0 +1,31 @@
+package com.google.sps.servlets;
+
+public class Course {
+  String name;
+  String professor;
+  Long units;
```
lqueipom, line 6: Why is this a Long and not an int?

ninaprabhu: Datastore stores numbers as Longs

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,279 @@
...
+  @AfterEach
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Mock HttpServletRequest request;
+
+  @Test
+  /* Ensure the correct filters are set when all search parameters are specified. */
+  public void makeFiltersAllParams() {
```
hailed748, line 62: MakeFilters_AllParamsSet sounds better, also luke said underscores are good for test names. Also test names are UpperCamelCase

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,279 @@
...
+    request = Mockito.mock(HttpServletRequest.class);
+
+    when(request.getParameter("courseName")).thenReturn("CS 105");
+    when(request.getParameter("profName")).thenReturn("Smith");
+    when(request.getParameter("term")).thenReturn("Spring 2020");
+    when(request.getParameter("units")).thenReturn("1,2,3");
+
+    List<Filter> filters = searchObject.getFilters(request);
+
+    List<Filter> expectedFilters = new ArrayList<>();
```
hailed748, line 73: Move this above the line where you call searchObject.getFilters() to follow the initialize, action, assertion rule Travis mentioned

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,279 @@
...
+    expectedFilters.add(new FilterPredicate("professor", FilterOperator.EQUAL, "Smith"));
+    expectedFilters.add(new FilterPredicate("term", FilterOperator.EQUAL, "Spring 2020"));
+    expectedFilters.add(new FilterPredicate("units", FilterOperator.IN, units));
+
+    assertEquals(filters, expectedFilters);
+  }
+
+  @Test
+  /* Ensure no filters are made given no search parameters are specified. */
+  public void makeFiltersNoParams() {
```
hailed748, line 88: nit: MakeFilters_NoParamsSet

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+    request = Mockito.mock(HttpServletRequest.class);
+
+    // TODO: Find a better way to replicate null parameters -
+    // getting NullPointerException when trying other methods.
+    when(request.getParameter("courseName")).thenReturn("");
```
hailed748, line 94: make a function to make the request. I made one in my test file. Pretty sure it should for for you so feel free to copy it

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,279 @@
...
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    request = Mockito.mock(HttpServletRequest.class);
+
+    when(request.getParameter("courseName")).thenReturn("CS 105");
+    when(request.getParameter("profName")).thenReturn("Smith");
+    when(request.getParameter("term")).thenReturn("Spring 2020");
+    when(request.getParameter("units")).thenReturn("1,2,3");
+
+    List<Entity> expectedResults = new ArrayList<>();
+    Entity entity1 = new Entity("Course");
```
hailed748, line 119: probably make a function to create entities as well because it seems like you make a lot of them

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  @Test
+  /* Ensure the correct results are retrieved from Datastore given no search filters. */
+  public void GetResults_NoFiltersSet() {
+
+    request = createRequest(request, "", "", "select", "");
+    List<Entity> expectedResults = new ArrayList<>();
+
+    addCourseEntityWithList("CS 105", "Smith", "Spring 2020", 1, expectedResults, true);
```
lqueipom, line 117: What does this variable name mean here?

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+  private void assertEqualsCourseArrays(List<Course> courses, List<Course> expectedCourses) {
+    assertEquals(courses.size(), expectedCourses.size());
+    for (int i = 0; i < courses.size(); i++) {
+      Course actual = courses.get(i);
+      Course expected = expectedCourses.get(i);
+      assertEquals(actual.name, expected.name);
+      assertEquals(actual.professor, expected.professor);
+      assertEquals(actual.units, expected.units);
+      assertEquals(actual.term, expected.term);
```
lqueipom, line 213: Why are these actual.term? Is actual the Course object we are not using anymore? (or are you still using it for your own code)

ninaprabhu: I'm still using the Course object, so I don't have to change as much of my code (and it helps simplify it a tiny bit for me)

src/test/java/com/google/sps/servlets/SearchServletTest.java
ninaprabhu: Sorry, what variable? If you look at the function declaration I think it's name, professor, term, units, array to potentially add to, and whether to add.

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,216 @@
...
+    Entity entity = new Entity("Course");
+    entity.setProperty("name", name);
+    entity.setProperty("professor", professor);
+    entity.setProperty("units", units);
+    entity.setProperty("term", term);
+    db.put(entity);
+  }
+
+  private HttpServletRequest createRequest(
+      HttpServletRequest request, String name, String professor, String term, String units) {
```
LukeGMTravis, line 196: Nice use of a helper function! One critique though: when you call this function you end up entering a sequence of string params. This is a place where it would be really easy for someone to mix up the param order and malform the function call. One solution to this is to use an object so the params are mapped. A possible workaround is to comment in the function call what the params are (since it's not obvious):
```request = createRequest(request, /** name */ "", /** professor */ "", /** term */ "select", /** units */ "");```. Since this is a test I don't think this needs to be a big deal but I think using this workaround at least would be a good idea.

File: src/test/java/com/google/sps/servlets/SearchServletTest.java [hunk 32]
```diff
@@ -0,0 +1,216 @@
...
+  @AfterEach
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Mock HttpServletRequest request;
+
+  @Test
+  /* Ensure the correct filters are set when all search parameters are specified. */
+  public void MakeFilters_AllParamsSet() {
```
LukeGMTravis, line 63: Ideally when testing we want to treat the code we're testing as a black box. It was inputs and outputs but we don't really care what happens in between, we just want to ensure that a given set of inputs results in the correct outputs. To that end, unless this is code that's common between multiple endpoints, you don't need to specifically test `getFilters` and `getResults` (in face, and I should have caught this earlier, these functions should be private if they aren't meant to be shared elsewhere).

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+    addCourseEntity("CS 105", "Smith", "Spring 2020", 1);
+    expectedCourses.add(new Course("CS 105", "Smith", Long.valueOf(1), "Spring 2020"));
+
+    addCourseEntity("CS 105", "Smith", "Spring 2020", 2);
+    expectedCourses.add(new Course("CS 105", "Smith", Long.valueOf(2), "Spring 2020"));
+
+    addCourseEntity("CS 106", "Smith", "Spring 2020", 3);
+
+    List<Filter> filters = searchObject.getFilters(request); // Tested in makeFiltersAllParams.
```
LukeGMTravis, line 141: Ideally this is tested just via `doGet` since that should likely be the input to the SearchServlet "black box".

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,134 @@
...
+      String term = (String) entity.getProperty("term");
+      Course course = new Course(name, professor, numUnits, term);
+      courses.add(course);
+    }
+    return courses;
+  }
+
+  /* Store course. */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
LukeGMTravis, line 119: Ideally this should also be tested. That being said, since it's more for admins and not the end user I think it's alright to skip at this early stage. In a real world situation though you'd definitely want this to be tested too.

ninaprabhu: Since we're planning on using Dagm's version of this and he's tested it, I figured it wasn't relevant - even if it's only used by me in this stage (and will be deleted after the merge/pull) should we still test ideally?

LukeGMTravis: Ah, in that case since this code will never be live in production then no need to worry I'd say

src/test/java/com/google/sps/servlets/SearchServletTest.java
ninaprabhu: Makes sense, I just made a getHelper function that essentially does everything up to the converting into json and tested that - is that sufficient, or would we want to test the json conversion as well?

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+  @Mock HttpServletRequest request;
+
+  @Test
+  /* Ensure the whole doGet process works given all search filters are set. */
+  public void GetCourses_AllParamsSet() throws IOException {
+
+    request =
+        createRequest(
+            request, /*name*/
```
LukeGMTravis, line 64: Comments should have spaces in between the asterisks: https://google.github.io/styleguide/javaguide.html#s4.8.6-comments. That _might_ be why prettier didn't format them too well.

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,151 @@
...
+    addCourseEntity("CS 105", "Smith", "Spring 2020", 1);
+    expectedCourses.add(new Course("CS 105", "Smith", Long.valueOf(1), "Spring 2020"));
+
+    addCourseEntity("CS 105", "Smith", "Spring 2020", 2);
+    expectedCourses.add(new Course("CS 105", "Smith", Long.valueOf(2), "Spring 2020"));
+
+    addCourseEntity("CS 106", "Smith", "Spring 2020", 3);
+
+    List<Course> courses = searchObject.getHelper(request);
+    System.out.println(response);
```
LukeGMTravis, line 80: Left over

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+    addCourseEntity("CS 105", "Smith", "Spring 2020", 1);
+    expectedCourses.add(new Course("CS 105", "Smith", Long.valueOf(1), "Spring 2020"));
+
+    addCourseEntity("CS 105", "Smith", "Spring 2020", 2);
+    expectedCourses.add(new Course("CS 105", "Smith", Long.valueOf(2), "Spring 2020"));
+
+    addCourseEntity("CS 106", "Smith", "Spring 2020", 3);
+
+    List<Course> courses = searchObject.getHelper(request);
```
LukeGMTravis, line 79: Ideally you only want to test your class's public API. In this case that would be `doGet` (and `doPost`). You only want to resort to making a helper function like this public for testing if there really isn't a way to directly test your public API.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,140 @@
...
+  /* Create list of courses given request. Used for testing. */
+  public List<Course> getHelper(HttpServletRequest request) {
+    List<Filter> filters = getFilters(request);
+    List<Entity> results = getResults(filters);
+    List<Course> courses = getCourses(results);
+    return courses;
+  }
+
+  /* Create list of filters given parameters specified in request. */
+  public List<Filter> getFilters(HttpServletRequest request) {
```
LukeGMTravis, line 59: Looks like this should be private, same for several other methods here

ninaprabhu: When I try to test any function that calls these helper functions (like doGet or getHelper), I get an error if I don't make them public because they're not accessible - am I calling these functions incorrectly or anything like that?

LukeGMTravis: If you just call `doGet` directly in your test you get an error if, say, `getFilters` isn't public?

ninaprabhu: yes - it says something along the lines of "getFilters is private so it can't be accessed outside of this file"

LukeGMTravis: That seems to imply that you're trying to call `getFilters` directly somewhere, is that the case?

ninaprabhu: No :/

LukeGMTravis: Which line is the error from?

src/test/java/com/google/sps/servlets/SearchServletTest.java
ninaprabhu: Hmmm dang okay, I'm running into a weird NullPointerException with my response when trying to test doGet directly - apparently Dagm and Laura also ran into the same issue and talked to Travis about it and are just kinda doing the same thing with the helper function that does the same thing as doGet

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: For some reason, I think it got fixed now, so that's good!

src/test/java/com/google/sps/servlets/SearchServletTest.java
Tetheta: Yeah the Java strict typing makes this a bit of a nightmare. Dagm and I tried a few things then I told him to just use the helper, because it wasn't worth wasting hours on. 
Good enough to know that typically you want to test the public API

LukeGMTravis: Ah, if that's the way things are then it can't be helped, carry on!

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -0,0 +1,152 @@
...
+    assertEqualsCourseArrays(courses, expectedCourses);
+  }
+
+  @Test
+  /* Ensure the whole doGet process works as expected given no search filters. */
+  public void GetCourses_NoParamsSet() {
+
+    request =
+        createRequest(request, /*name*/ "", /*professor*/ "", /*term*/ "select", /*units*/ "");
```
LukeGMTravis, line 91: Same spacing here

PR #5, opened 2020-07-23
Add School Info Feature

hailed748:
Purpose:  creates as add school information feature that can be used to populate a database with information regarding a school when members of the school first use the app

Adds: One JavaServlet and one HTML page. The HTML page is temporary as this form can be move to another page in the future

Modifies: This feature is independent of all other features and does not modifying and existing files.

Screenshot 2020-07-24 at 11 50 02 AM


src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,171 @@
...
+
+    response.setContentType("text/html; charset=UTF-8");
+    response.setCharacterEncoding("UTF-8");
+    response.sendRedirect("/addSchoolInfo.html");
+  }
+  /**
+   * @return the request parameter, or the default value if the parameter was not specified by the
+   *     client
+   */
+  private String getParameter(HttpServletRequest request, String name, String defaultValue) {
```
lqueipom, line 167: It doesn't look like you use this function, maybe delete itt?

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,171 @@
...
+
+          newTerm.setProperty("Professor", newProfessorKey);
+          datastore.put(newTerm);
+        }
+      } else {
+        // existing school --> existing course --> existing professor
+        Entity existingCourse = courseQuery.get(0);
+        existingCourseKey = existingCourse.getKey();
+
+        Entity newTerm = new Entity("Term", existingCourseKey);
```
lqueipom, line 128: This would be an existing term right? (consider changing the variable name maybe)

hailed748: It's actually a new term in all cases because we are assuming you can't have duplicate terms. I didn't make that check, but for safety I will change it such that if a term already exists we won't make a new one.

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,171 @@
...
+
+    if (newSchoolKey != null) {
+      Entity newCourse = new Entity("Course", newSchoolKey);
+      newCourse.setProperty("course-name", courseName);
+      newCourse.setProperty("Units", units);
+      newCourseKey = newCourse.getKey();
+      datastore.put(newCourse);
+
+      Entity newProfessor;
+      newProfessor = new Entity("Professor", existingSchoolKey);
```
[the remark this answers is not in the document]
hailed748: yup good catch!

src/main/java/com/google/sps/servlets/addSchoolInfo.java
lqueipom: Perfect! :)

src/main/java/com/google/sps/data/PostRequestObject.java
```diff
@@ -0,0 +1,38 @@
+package com.google.sps.data;
+
+public class PostRequestObject {
+  String school;
+  String course;
+  String term;
+  String units;
```
ninaprabhu, line 7: int?

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+      if (courseQuery.size() == 0) {
+        Entity newCourse;
+        if (existingSchoolKey != null) {
+          newCourse = new Entity("Course", existingSchoolKey);
+        } else {
+          newCourse = new Entity("Course", newSchoolKey);
+        }
+
+        newCourse.setProperty("course-name", courseName);
+        newCourse.setProperty("Units", units);
```
ninaprabhu, line 94: maybe standardize this? so "Course Name" and "Units" or "course-name" and "units" so it's easier for Laura and I to use your database without having to manually refer to each property name

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+          System.out.println("OLD SCHOOL NEW COURSE NEW PROFESSOR");
+          Entity newProfessor;
+          newProfessor = new Entity("Professor", existingSchoolKey);
+          newProfessor.setProperty("professor-name", profName);
+          newProfessorKey = newProfessor.getKey();
+          datastore.put(newProfessor);
+
+          newTerm.setProperty("Professor", newProfessorKey);
+          datastore.put(newTerm);
+
```
ninaprabhu, line 118: Nit: Extra line

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+      datastore.put(newTerm);
+
+    } else {
+      Filter courseFilter = new FilterPredicate("course-name", FilterOperator.EQUAL, courseName);
+      Query courseQ = new Query("Course").setAncestor(existingSchoolKey).setFilter(courseFilter);
+      List<Entity> courseQuery =
+          datastore.prepare(courseQ).asList(FetchOptions.Builder.withDefaults());
+
+      // existing school --> new course
+      if (courseQuery.size() == 0) {
```
ninaprabhu, line 85: use .isEmpty()

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+
+        Filter professorFilter =
+            new FilterPredicate("professor-name", FilterOperator.EQUAL, profName);
+        Query professorQ =
+            new Query("Professor").setAncestor(existingSchoolKey).setFilter(professorFilter);
+        List<Entity> professorQuery =
+            datastore.prepare(professorQ).asList(FetchOptions.Builder.withDefaults());
+
+        // existing school --> new course --> new professor
+        if (professorQuery.size() == 0) {
```
ninaprabhu, line 108: .isEmpty()

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+          System.out.println("OLD SCHOOL NEW COURSE NEW PROFESSOR");
```
ninaprabhu, line 109: if this is just for debugging, remove

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+          Entity newProfessor;
+          newProfessor = new Entity("Professor", existingSchoolKey);
```
ninaprabhu, line 111: Why is this not Entity newProfessor = newEntity("Professor", existingSchoolKey)? I might be missing something but just wondering

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+
+        Filter professorFilter =
+            new FilterPredicate("professor-name", FilterOperator.EQUAL, profName);
+        Query professorQ =
+            new Query("Professor").setAncestor(existingSchoolKey).setFilter(professorFilter);
+        List<Entity> professorQuery =
+            datastore.prepare(professorQ).asList(FetchOptions.Builder.withDefaults());
+
+        // existing school --> existing course --> new professor
+        if (professorQuery.size() == 0) {
```
ninaprabhu, line 143: .isEmpty()

src/main/java/com/google/sps/servlets/addSchoolInfo.java
```diff
@@ -0,0 +1,179 @@
...
+          Entity newProfessor;
+          newProfessor = new Entity("Professor", existingSchoolKey);
+          newProfessor.setProperty("professor-name", profName);
+          newProfessorKey = newProfessor.getKey();
+          datastore.put(newProfessor);
+
+          newTerm.setProperty("Professor", newProfessorKey);
+          datastore.put(newTerm);
+
```
ninaprabhu, line 152: Nit: Remove line

hailed748:
I made everything lower case for entity names are lower case and properties are lower case as well, hopefully this will standardize things throughout.

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
```diff
@@ -0,0 +1,159 @@
...
+      Entity newProfessor = new Entity("professor", newSchoolKey);
+      newProfessor.setProperty("professor-name", profName);
+      newProfessorKey = newProfessor.getKey();
+      datastore.put(newProfessor);
+
+      Entity newTerm = new Entity("term", newCourseKey);
+      newTerm.setProperty("term", term);
+      newTerm.setProperty("professor", newProfessorKey);
+      datastore.put(newTerm);
+
```
ninaprabhu, line 74: Nit: extra line? Feel free to ignore this if this is in the style guide - in that case my bad haha

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
```diff
@@ -0,0 +1,159 @@
...
+        Filter professorFilter =
+            new FilterPredicate("professor-name", FilterOperator.EQUAL, profName);
+        Query professorQ =
+            new Query("professor").setAncestor(existingSchoolKey).setFilter(professorFilter);
+        List<Entity> professorQuery =
+            datastore.prepare(professorQ).asList(FetchOptions.Builder.withDefaults());
+
+        // existing school --> existing course --> new professor
+        if (professorQuery.size() == 0) {
```
ninaprabhu, line 138: .isEmpty()

src/main/webapp/addSchoolInfo.html
```diff
@@ -0,0 +1,29 @@
...
+    <link rel="stylesheet" href="style.css" />
+  </head>
+  <body>
+    <p>Add A Course Below</p>
+    <form action="/add-info" method="POST">
+        <p>School Name</p>
+        <textarea name="school-name">SchoolName</textarea>
+        <br><br>
+        <p>Course Name</p>
+        <textarea name="course-name">CoureName</textarea>
```
ninaprabhu, line 15: Nit: typo

src/main/webapp/addSchoolInfo.html
```diff
@@ -0,0 +1,29 @@
...
+        <br><br>
+        <p>Term</p>
+        <textarea name="term">Term</textarea>
+        <br><br>
+        <p>Unit</p>
+        <textarea name="units">Units</textarea>
```
ninaprabhu, line 21: Sidenote: I've been using checkboxes instead of textareas for my units section and I think that makes it easier in the long run (since you won't run into people putting in invalid numbers), maybe consider changing to that?

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
hailed748: Oh I use an extra line to separate each entity. I think its allowed, kinda a grey area

src/main/webapp/addSchoolInfo.html
hailed748: I changed it to a number input box so they can only input a number from 1 - 24

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
lqueipom: Agree with Dagmawi on this, makes it a lot easier to read :)

src/test/java/com/google/sps/AddingSchoolInfoTest.java
```diff
@@ -0,0 +1,54 @@
...
+//   public void processPostRequestShouldPopulateObject() {
+
+//     PostRequestObject testPostObject = AddSchoolInfo.proccesPostRequest(MockPostObject);
+//     Assert.assertEquals(testPostObject.getSchool(), mockPost.getParameter("school-name"));
+//     Assert.assertEquals(testPostObject.getCourse(), mockPost.getParameter("course-name"));
+//     Assert.assertEquals(testPostObject.getTerm(), mockPost.getParameter("term"));
+//     Assert.assertEquals(testPostObject.getUnits(), mockPost.getParameter("units"));
+//     Assert.assertEquals(testPostObject.getProf(),mockPost.getParameter("professor-name"));
+//   }
+// }
```
lqueipom, line 54: Wait, why is this file all commented out?

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
```diff
@@ -0,0 +1,159 @@
...
+  private Key newProfessorKey;
+  private Key newTermKey;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    String schoolName = request.getParameter("school-name");
+    String courseName = request.getParameter("course-name");
+    String term = request.getParameter("term");
+    Integer units = Integer.parseInt(request.getParameter("units"));
```
lqueipom, line 37: Don't you have to say "int" as type instead of "Integer"? (I remember that I tried to run with Integer and for some reason told me it did not exist)

src/test/java/com/google/sps/AddingSchoolInfoTest.java
hailed748: I need to delete this file I moved it to a different branch

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
hailed748: I have actually always been using Integer this entire internship lol it seems to work

lqueipom: JAJAJA perfect, good that it works for youu :)

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
ninaprabhu: Sounds good :)

Tetheta:
Starting review

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
```diff
@@ -0,0 +1,159 @@
...
+public class AddSchoolInfo extends HttpServlet {
+
+  private Key existingSchoolKey;
+  private Key existingCourseKey;
+  private Key existingProfessorKey;
+
+  private Key newSchoolKey;
+  private Key newCourseKey;
+  private Key newProfessorKey;
+  private Key newTermKey;
```
Tetheta, line 29: Do these actually need to live for the duration of this Servlet? 
Typically I'd advise keeping parameters close to where they're used and as locally scoped as possible.

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
Tetheta: I also prefer the separate line, clarity is most important here. The main thing we don't like it two separate lines, or a separate line after opening a block via {

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
```diff
@@ -0,0 +1,159 @@
...
+    String profName = request.getParameter("professor-name");
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    Filter schoolFilter = new FilterPredicate("school-name", FilterOperator.EQUAL, schoolName);
+    Query schoolQ = new Query("school").setFilter(schoolFilter);
+    List<Entity> schoolQuery =
+        datastore.prepare(schoolQ).asList(FetchOptions.Builder.withDefaults());
+
+    if (schoolQuery.isEmpty()) {
```
Tetheta, line 47: A lot of these if/else statements could be reduced to a function call that encapsulates that switching logic.

This one for instance could be: `getOrCreateSchoolKey(schoolQuery)` and return the school key. Then you've reduced 10 lines into 1 and it's easier to reason about.

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
```diff
@@ -0,0 +1,159 @@
...
+
+      Entity newProfessor = new Entity("professor", newSchoolKey);
+      newProfessor.setProperty("professor-name", profName);
+      newProfessorKey = newProfessor.getKey();
+      datastore.put(newProfessor);
+
+      Entity newTerm = new Entity("term", newCourseKey);
+      newTerm.setProperty("term", term);
+      newTerm.setProperty("professor", newProfessorKey);
+      datastore.put(newTerm);
```
Tetheta, line 73: You're handling everything at once it seems (course, professor, and term).
While this works, it means there's a lot of code to traverse to get to the details of how the other term creations for example are used and it's easy to miss an edge case.

I'd advise splitting this up into three functions: `getOrCreateCourse`, `getOrCreateProfessor`, and `getOrCreateTerm`.
Then create some booleans if needed like `isNewSchool`, `isNewProfessor`, `isNewTerm` etc. With corresponding functions as needed.

Have the functions pass back the Entity created, then use that for your final data store puts at the end.

Having a conceptual:
- Create course
- Create professor
- Create term
- Store them

With necessary logic aggregated as you go via booleans/etc I think would be clearer.

It's also possible I'm misunderstanding this and you need this extra complexity/branching. If so ping me and we can chat about it further.

src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
+  <threadsafe>false</threadsafe>
+  <sessions-enabled>true</sessions-enabled>
+  <runtime>java8</runtime>
+  <static-files>
+    <!-- prevent unwanted caching when accessing via the web preview server -->
+    <include path="/**" expiration="0s" />
```
Tetheta, line 8: Does this only affect the preview server or will it affect prod as well? Not like it matters here, just curious.

hailed748: Not really sure how this works. This was a boilerplate given to us in walk-through.

hailed748:
Adds: refactored servlet now called AddSchoolData, unit test cases under AddSchoolDataTest

Modifies: pom.xml

Removed: Login and Data servlet (these were causing build problems because I was missing some dependencies that Laura and Nina used. I figured i would delete them for now because they will be in the master)

Your suggestion on breaking it up was really good, I tried my best to make it simpler. However I couldn't figure out how to simplify the addSchoolData() function because I need to pass in the parent keys upon entity creation; there is no way to add the parent after creation. Let me know if you have ideas to simplify this.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,133 @@
...
+
+  public List<Entity> findQueryMatch(
+      DatastoreService db, String entityType, String entityProperty, String propertyValue) {
+    Filter filter = new FilterPredicate(entityProperty, FilterOperator.EQUAL, propertyValue);
+    Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    return result;
+  }
+
+  public void isNewFlagSwitcher(
```
lqueipom, line 48: I am having trouble figuring out what the function does by the name, maybe consider changing it?

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    List<Entity> queryProfessor = findQueryMatch(db, "professor", "professor-name", profName);
+    if (!queryProfessor.isEmpty()) {
+      existingSchoolKey = queryProfessor.get(0).getKey();
+      isNewProfessor = false;
+    }
+  }
+
+  public Entity createSchool(DatastoreService db, String schoolName) {
```
lqueipom, line 69: If you initialize datastore at the beginning, do you still need to pass in db as a param?

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,133 @@
...
+    Entity newProfessor = new Entity("professor", parent);
+    newProfessor.setProperty("professor-name", name);
+    db.put(newProfessor);
+    return newProfessor;
+  }
+
+  public Entity createTerm(DatastoreService db, String term, Key professor, Key parent) {
+    Entity newTerm = new Entity("term", parent);
+    newTerm.setProperty("term", term);
+    newTerm.setProperty("professor", professor);
```
lqueipom, line 94: It would be good if that variable was called professorKey

pom.xml
```diff
@@ -66,7 +107,7 @@
         <version>2.2.0</version>
         <configuration>
           <!-- TODO: set project ID. -->
-          <deploy.projectId>nina-laura-dagm-step-2020</deploy.projectId>
+          <deploy.projectId>dagmawi-step-2020</deploy.projectId>
```
ninaprabhu, line 110: Does our project ID not work for your testing? Just curious

src/main/webapp/AddSchoolData.html
```diff
@@ -0,0 +1,29 @@
...
+  </head>
+  <body>
+    <p>Add A Course Below</p>
+    <form action="/add-info" method="POST">
+        <p>School Name</p>
+        <textarea name="school-name">SchoolName</textarea>
+        <br><br>
+        <p>Course Name</p>
+        <textarea name="course-name">CourseName</textarea>
+        <br><br>
```
ninaprabhu, line 16: Just a small note: I found it more intuitive to leave these text areas blank and give them labels, so we don't have to delete the "SchoolName", "CourseName", existing info every time before adding a course. Not something you need to do, just figured I'd mention that in case you found it easier to work with!

pom.xml
hailed748: I think when I pulled it reverted it

src/main/webapp/AddSchoolData.html
```diff
@@ -0,0 +1,29 @@
...
+        <p>Term</p>
+        <textarea name="term">Term</textarea>
+        <br><br>
+        <p>Units</p>
+        <input type="number" name="units" min="1" max="24">
+        <br><br>
```
lqueipom, line 22: Consider switching out your bracket tags for padding and margin in CSS (Luke made a comment of this in my PR)

.gitignore
```diff
@@ -1,6 +1,3 @@
-<<<<<<< HEAD
-target/
-=======
-# Excludes libraries when commiting. 
-target/
->>>>>>> a18dee48c295115865c85e4418dfb57c05a27f4f
+target/*
+node_modules/*f
```
Tetheta, line 2: Why *f?

pom.xml
Tetheta: Please change this.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private Boolean isNewSchool = true;
+  private Boolean isNewCourse = true;
+  private Boolean isNewProfessor = true;
+
+  private Key existingSchoolKey;
+  private Key existingCourseKey;
+  private Key existingProfessorKey;
+
+  private DatastoreService db = DatastoreServiceFactory.getDatastoreService();
```
Tetheta, line 30: Same question on the lifecycle of these. Typically you want very few things to live outside of your post/etc function.

Having things live longer can cause issues if you don't reset them all and those bugs are very hard to track down.

src/main/java/com/google/sps/servlets/AddSchoolInfo.java
hailed748: I was able to get rid of the new...Key variables but I couldn't figure out how to get rid of the existing key variables

hailed748:
@hailed748 ping test?

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public Entity createTerm(String term, Key professor, Key parent) {
+    Entity newTerm = new Entity("Term", parent);
+    newTerm.setProperty("term", term);
+    newTerm.setProperty("professorKey", professor);
+    db.put(newTerm);
+    return newTerm;
+  }
+
+  public void addSchoolData(DatastoreService db, HttpServletRequest request) {
```
Tetheta, line 98: Typically we'll put helper functions at the bottom, and the main functions near the top.

That way you can read top to bottom and move as little as possible to see where the code goes.

Also, these should all be private functions.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,132 @@
...
+    String schoolName = request.getParameter("school-name");
+    String courseName = request.getParameter("course-name");
+    String termName = request.getParameter("term");
+    String profName = request.getParameter("professor-name");
+    Long units = Long.parseLong(request.getParameter("units"));
+
+    isNewEntityDetector(schoolName, courseName, profName);
```
Tetheta, line 105: I'd turn this into three small functions that return booleans. That way you don't have to use global variables.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,132 @@
...
+
+    if (isNewSchool) {
+      Entity school = createSchool(schoolName);
```
Tetheta, line 108: You don't use these variables for anything, so there's no need to have them.

I'd either advise having your create functions not return anything (since it's not needed), or have them return entities and do a single datastore.put at the end of the entire function. The former is easier here, the latter has the advantage of being very clear about where the storage operation is taking place.

src/main/webapp/WEB-INF/appengine-web.xml
Tetheta: Interesting, guess it doesn't matter.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,132 @@
...
+  private Boolean isNewCourse = true;
+  private Boolean isNewProfessor = true;
+
+  private Key existingSchoolKey;
+  private Key existingCourseKey;
+  private Key existingProfessorKey;
+
+  private DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+
+  @Override
```
Tetheta, line 32: Nice job cleaning this up. It feels a lot more readable now, and hopefully it's also easier to test.
Just a few things and then should be good to go.

.gitignore
hailed748: typo

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
```diff
@@ -0,0 +1,449 @@
...
+  @Test
+  public void AddingNewSchool_CreatesNewEntityForAll() {
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    request = Mockito.mock(HttpServletRequest.class);
+    when(request.getParameter("school-name")).thenReturn("MIT");
+    when(request.getParameter("course-name")).thenReturn("6.006");
+    when(request.getParameter("term")).thenReturn("Spring 2020");
+    when(request.getParameter("units")).thenReturn("12");
+    when(request.getParameter("professor-name")).thenReturn("Jason Ku");
+    newSchoolObject.addSchoolData(db, request);
```
Tetheta, line 73: This is used a ton, perfect candidate for creating a helper function.

Something along the lines of `createRequest(schoolName, termName, ...)` 
I think calling out that the db is created and this is added to the datastore is good per test, but repeating these details every time is just noisy.

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
```diff
@@ -0,0 +1,449 @@
...
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Mock HttpServletRequest request;
+
+  @Mock HttpServletRequest requestB;
+
+  @Test
+  public void AddingNewSchool_CreatesNewEntityForAll() {
```
Tetheta, line 65: Something we forgot to talk about that would help me a lot when reviewing this.

Typically for each test you'll have three sections: Setup, Act, and Assert.

It may even be useful to call them out explicitly via comment, though typically we just put a blank line between.

What this means for your code here:

Setup is creating the request mock, and any variables needed to check against later (like expected entities/lengths/etc)

Act is actually doing the action, in this case just calling the function on your AddSchoolInfo class.

Assert is anything with assertEquals/etc, where you're verifying the behavior.

You have all the right stuff here, it'd just be easier to read if it was moved into that order.

Tetheta: Also sometimes you might have a pre-assert section where you make sure the DB is empty/etc. Though in this case since you're mocking it out each time I don't think that's necessary.

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
```diff
@@ -0,0 +1,449 @@
...
+
+    assertEquals(2, children.size());
+  }
+
+  public List<Entity> findQueryMatch(
+      DatastoreService db,
+      String entityType,
+      String entityProperty,
+      String propertyValue,
+      Boolean onlyType) {
```
Tetheta, line 431: If this boolean is true, do you nee entityProperty and propertyValue? Seems like you could make those both optional parameters here, or overload your function here for the two different query patterns.

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
```diff
@@ -0,0 +1,449 @@
...
+    Key school =
+        findQueryMatch(db, "School", "school-name", request.getParameter("school-name"), false)
+            .get(0)
+            .getKey();
+    List<Entity> children = findChildren(db, "Professor", school);
+
+    assertEquals(2, children.size());
+  }
+
+  public List<Entity> findQueryMatch(
```
Tetheta, line 426: This doesn't need to be public does it?

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
```diff
@@ -0,0 +1,449 @@
...
+import javax.servlet.http.HttpServletRequest;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+/** */
+public final class AddSchoolDataTest {
```
Tetheta, line 41: Things I don't see covered yet:

1. What if the data is malformed or missing in the request?
2. What if the datastore.put throws an error?
3. What if the requests are identical?

It's not necessary to handle all these yet, but adding a TODO comment to handle them in the future is typically useful.

src/main/webapp/index.html
```diff
@@ -9,6 +9,7 @@
       src="https://www.gstatic.com/charts/loader.js"
     ></script>
     <script src="script.js"></script>
+    <script src="Graphscript.js"></script>
```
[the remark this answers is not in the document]
hailed748: I just changed this. I think you began the review before my other commits went through.

src/main/webapp/Graphscript.js
```diff
@@ -0,0 +1,87 @@
...
+  liElement.innerText = text;
+  return liElement;
+}
+
+// This is a temporary function, once I connect with Nina will get
+// better way to create/access class ids.
+function getClassID() {
+  var classId = Math.floor(Math.random() * 101);
+  document.getElementById("class-id").innerHTML = classId;
+}
```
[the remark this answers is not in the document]
hailed748: I believe this is Nina's code that got pulled when I pulled from master. Likewise with anything course/login related

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -0,0 +1,157 @@
...
+    Entity course;
+    Entity professor;
+    Entity term;
+
+    if (isNewSchool) {
+      school = createSchool(schoolName);
+      course = createCourse(courseName, units, school.getKey());
+      professor = createProfessor(profName, school.getKey());
+      term = createTerm(termName, professor.getKey(), course.getKey());
+      db.put(school);
```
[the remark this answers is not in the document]
hailed748: I wasn't quite able to easily create a list so I just went with option 1 and created a function. I will look more into how to easily add multiple things to a list to get the second version working.

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
hailed748: I throw an exception now if there is an error on the .put(), and i throw an illegal argument exception if any malformed data is being put. I do not know how to make tests for these though.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -152,6 +140,28 @@ private Entity createTerm(String term, Key professor, Key parent) {
...
+      }
+      if (term != null) {
+        db.put(term);
+      }
+      if (professor != null) {
+        db.put(professor);
+      }
+    } catch (DatastoreFailureException | IllegalArgumentException e) {
+      throw e;
+    }
```
Tetheta, line 165: Nice job, another advantage of co-locating the puts is you can do something like this.

src/main/webapp/graphScript.js
```diff
@@ -0,0 +1,34 @@
...
+          histogram: {
+            hideBucketItems: true
+          }
+        };
+        const hourChart = new google.visualization.Histogram(document.getElementById("hours-chart"));
+        hourChart.draw(hourData, hourOptions);
+
+      const difficultyData = new google.visualization.arrayToDataTable(data.difficulty);
+      const diffOptions = {
+          title: "Diffculty of Class",
```
Tetheta, line 23: nit: Typo, should be "Difficulty"

src/test/java/com/google/sps/servlets/AddSchoolDataTest.java
```diff
@@ -225,36 +171,20 @@ public void AddingNewTerm_NoNewSchool_NewCourse_NoNewProfessor() {
   @Test
   public void AddingNewTerm_NoNewSchool_NewCourse_NewProfessor() {
```
Tetheta, line 172: Just a nit, but one other thing handy to do in tests is make your variables/asserts/etc follow the same order as the test name (and make all test names follow the same order if you have a bunch of essentially boolean logic).

So in this case, declaring expectedTerm, then school, then course, then professor. Then querying & asserting in that same order.

PR #6, opened 2020-07-24
Attaching user ID to rating

lqueipom:
DataServlet:
1. Checked whether a given user had a key assigned to them. If not, it would generate a "User" entity where the user-name would be the user's email.
2. Querys in an ancestor path whether a rating exists that has the user key as a property.
3. If Query is empty, we will generate a new Rating entity. If not empty, we will change all values of the properties except that of the userKey. 

Index:
1. Added spaces for hours of work per week and difficulty of the class.


src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -50,9 +54,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get written feedback.
     String classFeedback = request.getParameter("class-input");
-    String professorFeedback = request.getParameter("prof-input");
-    // Get class/professor ratings.
     String classRating = request.getParameter("rating-class");
+    Int hoursOfWork = request.getParameter("hoursOfWork");
```
ninaprabhu, line 58: Nit: workHours or hoursWork instead?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,19 +97,67 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    List<Entity> userQueryList =
+        datastore.prepare(userQuery).asList(FetchOptions.Builder.withDefaults());
+
+    // User has not reviewed any rating.
+    if (userQueryList.size() == 0) {
+      Entity newReviewer = new Entity("User");
+      newReviewer.setProperty("user-email", userEmail);
+      Entity userID = newReviewer;
+    } else {
+      Entity userID = userQueryList.get(0);
```
ninaprabhu, line 116: Make sure this doesn't run into an error with userID not being defined outside of the if/else statement, I feel like Java is sometimes really weird about this kind of thing

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,19 +97,67 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    Query<Entity> reviewerQuery =
+        Query.newEntityQueryBuilder()
+            .setKind("Rating")
+            .setFilter(CompositeFilter.and(PropertyFilter.eq("reviewer-id", userID.getKey())))
+            .build();
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(commentsEntity);
-    datastore.put(professorEntity);
+    if (reviewerQuery != null) {
```
ninaprabhu, line 129: I think you'd have to do datastore.run(reviewerQuery) before you could see the results

src/main/webapp/index.html
```diff
@@ -47,6 +47,17 @@
...
+      <p>How many hours per week did you spend on this class?</p>
+      <input type="number" id="hoursOfWork" name="hoursOfWork" min="1" />
+      <label for="hoursOfWork">Hours (include in-class time)</label>
+
+      <br /><br />
+      <p>How difficult did you find this class?</p>
+      <input type="number" id="difficulty" name="difficulty" min="0" max="5" />
+      <label for="difficulty"
+        >Difficulty (From 0 = Very Easy to 5 = Very Hard)</label
+      >
```
ninaprabhu, line 59: I feel like this styling is a little bit weird with the >'s on different lines from their preceding tags?

lqueipom: I agree but for some reason prettier is formatting it like that. Will bring it up on the standup.

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: I did something similar I believe, lmk if you believe what I did works!

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -50,9 +54,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get written feedback.
     String classFeedback = request.getParameter("class-input");
-    String professorFeedback = request.getParameter("prof-input");
-    // Get class/professor ratings.
     String classRating = request.getParameter("rating-class");
+    Int workHours = request.getParameter("hoursOfWork");
+    Int difficulty = request.getParameter("difficulty");
+    String professorFeedback = request.getParameter("prof-input");
```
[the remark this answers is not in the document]
hailed748: Nina just mentioned in stand-up that Text objects sometimes leads to errors where querying so probably just set a limit on a comment to 1500 characters which matches the 1500 byte limit of a string object in Datastore.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -50,9 +54,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     String professorRating = request.getParameter("rating-professor");
```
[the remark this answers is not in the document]
lqueipom: It was more to put like name to the rating but wanted to know whether that is something we want to remove now that we have the sentiment analysis that creates the "rating" of the class in a sense? Or do we want to still allow them to rate excellent, etc and store it as ints?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,19 +97,70 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+          Entity.newBuilder(datastore.get(rankingKey))
+              .set("comments-professor", professorFeedback)
+              .set("score-professor", professorScore)
+              .set("perception", professorRating)
+              .set("difficulty", difficulty)
+              .set("hours", workHours)
+              .build();
+      datastore.update(rankingEntity);
+    } else {
+      Entity professorRatingEntity = new Entity("Rating");
```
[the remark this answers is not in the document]
lqueipom: Perfect! For now, I will assume that we will store the term in the url as "term=whateverTerm" so I can write the code that is supposed to do this.

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
lqueipom: Kept the rating but changed it to ints :)

lqueipom:
For some reason, target files are still being committed. Will look into that but asked for review so content could be checked :)

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,19 +100,80 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      userID = newReviewer;
+    } else {
+      userID = userQueryList.get(0);
+    }
+
+    currentTermKey = request.getParameter("term");
+    Entity currentTerm = datastore.get(currentTermKey);
+    currentProfKey = currentTerm.getProperty("Professor");
+
+    // Checks for term rating, only need to query one as prof and term get reviewed simultaneously.
```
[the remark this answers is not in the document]
LukeGMTravis: There's several concerns below that are somewhat related to this so I'll hold off on reviewing further for now until you've had a chance to address this.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,19 +100,80 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(commentsEntity);
-    datastore.put(professorEntity);
+    Filter userFilter = new FilterPredicate("user-email", FilterOperator.EQUAL, userEmail);
+    Query userQuery = new Query("User").setFilter(userFilter);
+    List<Entity> userQueryList =
+        datastore.prepare(userQuery).asList(FetchOptions.Builder.withDefaults());
+    Entity userID = new Entity("User");
+
+    // User has not reviewed any rating.
```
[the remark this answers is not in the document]
lqueipom: On her side, Nina used an API to fix this issue. So will connect with her to get it the userID through her.

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
lqueipom: Makes complete sense, sorry about that we should have checked this before. I will separate these ratings into two different types of entities.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,20 +101,83 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      Entity newReviewer = new Entity("User");
+      newReviewer.setProperty("user-email", userEmail);
+      userEntity = newReviewer;
+    } else {
+      userEntity = userQueryList.get(0);
+    }
 
+    // Has to be added to URL.
+    currentTermKey = request.getParameter("term");
+    currentSchoolKey = request.getParameter("school");
```
LukeGMTravis, line 127: Since you're using a full key for the term you shouldn't need a school key to help distinguish it. Similarly, you should be able to get the school from the user or at least the professor so you shouldn't need to pass it as a param anyway.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,20 +101,83 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+
+    // Checks for term rating by reviewer.
+    Filter termReviewerFilter =
+        new FilterPredicate("reviewer-id", FilterOperator.EQUAL, userEntity.getKey());
+    Query termRatingQuery =
+        new Query("Rating Term").setAncestor(currentTermKey).setFilter(termReviewerFilter);
+    List<Entity> termRatingQueryList =
+        datastore.prepare(termRatingQuery).asList(FetchOptions.Builder.withDefaults());
+
+    // Checks for professor rating by reviewer.
```
LukeGMTravis, line 139: If I'm understanding what's below correctly you want to create a professor rating and a term rating each time a rating is submitted via this endpoint? Why is that? Where do you envision this endpoint being called from? I would expect that if you submit a review on a professor's page you only create a professor review, and same for a term. Am I misunderstanding the spec?

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: Okay! Then, term should suffice for this 👍

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: Talked about this during 1:1, will implement a fix now!

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
LukeGMTravis: If this isn't being done in this PR can you add a TODO comment about it?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -92,20 +99,64 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+
+    // Checks for term rating by reviewer.
+    Filter termReviewerFilter =
+        new FilterPredicate("reviewer-id", FilterOperator.EQUAL, userEntity.getKey());
+    Query termRatingQuery =
+        new Query("Rating Term").setAncestor(currentTermKey).setFilter(termReviewerFilter);
+    List<Entity> termRatingQueryList =
+        datastore.prepare(termRatingQuery).asList(FetchOptions.Builder.withDefaults());
+
+    if (termRatingQueryList.isEmpty()) {
```
LukeGMTravis, line 135: This if-else has a lot of duplicated code, it looks like you can DRY it up

src/main/webapp/index.html
```diff
@@ -22,71 +22,33 @@
       <button onclick="getClassID()">Get class ID</button>
       <br /><br />
 
-      <textarea name="class-input">Write opinion on class here.</textarea>
+      <textarea name="class-input" maxlength="1500">Write opinion on class here.</textarea>
```
LukeGMTravis, line 25: This is good but keep in mind that users *can* get around this, you'll still need to handle a longer maxlength on the server. (Something to keep in mind, no change required just now.)

src/main/webapp/index.html
```diff
@@ -22,71 +22,33 @@
...
-      <label for="acceptable">Acceptable.</label>
-      <input type="radio" id="neutral" name="rating-class" value="neutral" />
-      <label for="neutral">Neutral.</label>
-      <input type="radio" id="not-good" name="rating-class" value="not-good" />
-      <label for="not-good">Not good.</label>
-      <input type="radio" id="terrible" name="rating-class" value="terrible" />
-      <label for="terrible">Terrible.</label>
+      <input type="rating-class" id="rating-class" name="rating-class" min="0" max="5" />
+      <label for="rating-class">Class Rating (From 0 = Terrible to 5 = Excellent)</label>
+      <br><br>
```
LukeGMTravis, line 31: Now that you're out of the prototype phase I'd recommend switching from using multiple br elements to adding padding/margin via css.

(Also your br tags are inconsistently styled.)

src/main/webapp/index.html
LukeGMTravis: Very odd... I have no idea > . <

Html will strip whitespace padding though so if you put the text inside the label on its own line that might work better and not get mutilated by prettier?

src/main/webapp/index.html
```diff
@@ -22,71 +22,33 @@
...
-      <input
-        type="radio"
-        id="terrible"
-        name="rating-professor"
-        value="terrible"
-      />
-      <label for="terrible">Terrible.</label>
+      <input type="rating-prof" id="rating-prof" name="rating-prof" min="0" max="5" />
+      <label for="rating-prof">Professor Rating (From 0 = Terrible to 5 = Excellent)</label>
+      <br><br>
```
LukeGMTravis, line 51: Same as above

src/main/webapp/index.html
```diff
@@ -22,71 +22,33 @@
...
-      <label for="terrible">Terrible.</label>
+      <input type="rating-class" id="rating-class" name="rating-class" min="0" max="5" />
+      <label for="rating-class">Class Rating (From 0 = Terrible to 5 = Excellent)</label>
+      <br><br>
+
+      <p>How many hours per week did you spend on this class?</p>
+      <input type="number" id="hoursOfWork" name="hoursOfWork" min="1" />
+      <label for="hoursOfWork">Hours (include in-class time)</label>
+
+      <br /><br />
```
LukeGMTravis, line 37: Same as above

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
LukeGMTravis: Hm, I'm not seeing a TODO comment...

src/main/java/com/google/sps/servlets/DataServlet.java
LukeGMTravis: Ah, nice and DRY : ) You could shorten this even more if you wanted with a ternary:
```Entity classRatingEntity = termRatingQueryList.isEmpty() ? new Entity("Rating Term", currentTermKey) : termRatingQueryList.get(0);```

Up to you if which way you like best for readability

File: src/test/java/com/google/sps/DataServletTest.java [hunk 17]
```diff
@@ -0,0 +1,229 @@
...
+    when(request.getParameter("rating-professor")).thenReturn("3");
+    when(request.getParameter("languages")).thenReturn("false");
+    when(request.getParameter("ID")).thenReturn("numberOneId");
+    when(request.getParameter("CourseClass"))
+        .thenReturn(Course("Spring 2020", "Smith", 12, termKey));
+
+    newTermRating.addTermRating(request);
+
+    Entity termRatingQuery =
+        queryEntities("Term Rating", reviewer - id, request.getParameter("reviewer-id")).get(0);
```
hailed748, line 131: missing quotes around "reviewer-id"

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,229 @@
...
+        expectedTermRatingEntity.getProperty("difficulty"),
+        termRatingQuery.getProperty("difficulty"));
+    expectedTermRatingEntity.setProperty("comments-professor", "The professor was still okay.");
+    assertEquals(
+        expectedTermRatingEntity.getProperty("comments-professor"),
+        termRatingQuery.getProperty("comments-professor"));
+    expectedTermRatingEntity.setProperty("perception-professor", 3);
+    assertEquals(
+        expectedTermRatingEntity.getProperty("perception-professor"),
+        termRatingQuery.getProperty("perception-professor"));
```
hailed748, line 159: Add spacing between the assertEquals, its hard to read

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,229 @@
...
+    when(request.getParameter("ID")).thenReturn("numberOneId");
+    Entity termEntity = new Entity("Term");
+    Key termKey = termEntity.getKey();
+    when(request.getParameter("CourseClass"))
+        .thenReturn(Course("Spring 2020", "Smith", 12, termKey));
+
+    newTermRating.addTermRating(request);
+
+    Entity termRatingQuery =
+        queryEntities("Term Rating", reviewer - id, request.getParameter("reviewer-id")).get(0);
```
hailed748, line 184: missing quotes around "reviewer - id" I think you copied this throughout so check all of them

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,229 @@
...
+    Entity expectedTermRatingEntity = new Entity("Term Rating");
```
hailed748, line 66: This is not necessarily related to testing. But I think you mentioned that Luke said to have a single rating entity that has the professor and term rating together, If we are doing this the Entity name could be simplified to "Rating" as all our other entity types are single words

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,229 @@
...
+    when(request.getParameter("ID")).thenReturn("numberOneId");
+    Entity termEntity = new Entity("Term");
+    Key termKey = termEntity.getKey();
+    when(request.getParameter("CourseClass"))
+        .thenReturn(Course("Spring 2020", "Smith", 12, termKey));
+
+    newTermRating.addTermRating(request);
+
+    Entity termRatingQuery =
+        queryEntities("Term Rating", reviewer - id, request.getParameter("reviewer-id")).get(0);
```
hailed748, line 65: should be "reviewer-id"

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,229 @@
...
+        expectedTermRatingEntity.getProperty("reviewer-id"),
+        termRatingQuery.getProperty("reviewer-id"));
+    expectedTermRatingEntity.setProperty("perception-class", 4);
+    assertEquals(
+        expectedTermRatingEntity.getProperty("perception-class"),
+        termRatingQuery.getProperty("perception-class"));
+    expectedTermRatingEntity.setProperty("hours", 8);
+    assertEquals(
+        expectedTermRatingEntity.getProperty("hours"), termRatingQuery.getProperty("hours"));
+    expectedTermRatingEntity.setProperty("difficulty", 4);
```
hailed748, line 82: You will likely get an error thrown here and on all number comparisons because datastore stores them as Longs and when you just put 4 I think its made an int. So you need to case the 4 as a Long

src/test/java/com/google/sps/DataServletTest.java
lqueipom: Sorry about this, you are completely right hahaha

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,243 @@
...
+        termRatingQuery.getProperty("score-professor"));
+  }
+
+  @Test
+  public void AddingTermRatingWithTranslation() {
+
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    request = Mockito.mock(HttpServletRequest.class);
+
+    when(request.getParameter("class-input")).thenReturn("Esta clase es muy interesante.");
```
LukeGMTravis, line 215: This is probably alright for now but it's not unlikely that this test will break later since you're relying on the translation service to always behave the same way (which I doubt it will in all cases). More ideal is to mock this out since it's not directly under test. (You should usually avoid mocking things you don't need to but since this is technically non-deterministic as far as you're concerned then it makes since in this case.)

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,243 @@
...
+    when(request.getParameter("rating-class")).thenReturn("4");
```
LukeGMTravis, line 216: If this isn't being tested then it's best not to set it. Same below

src/test/java/com/google/sps/DataServletTest.java
lqueipom: Oh that makes sense, I will look into it but when you mean "mock this out" in terms of translation what specifically are you referring to?

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,222 @@
...
+  @AfterEach
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Mock HttpServletRequest request;
+
+  @Test
+  public void AddingNewTermRating_NoTranslation() {
+
```
LukeGMTravis, line 42: Nit: extra line

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,222 @@
...
+    Key termKey = termEntity.getKey();
+    when(request.getParameter("Course"))
+        .thenReturn(Course("Spring 2020", "Smith", 12, termKey));
+
+    newTermRating.addTermRating(request);
+
+    Entity termRatingQuery =
+        queryEntities("Rating", "reviewer-id", request.getParameter("reviewer-id")).get(0);
+    Entity expectedTermRatingEntity = new Entity("Rating");
+    expectedTermRatingEntity.setProperty("comments-term", "I do not like this.");
```
LukeGMTravis, line 65: This would be easier to follow if you set it right next to your mock response. Then I would know at a glance that they're the same

src/test/java/com/google/sps/DataServletTest.java
LukeGMTravis: (same below)

src/test/java/com/google/sps/DataServletTest.java
LukeGMTravis: It's also a bit weird creating an expectedTermRatingEntity when you go on to just check each value individually. Probably simpler to just check the value directly if you can't compare actual entities?

LukeGMTravis: Speaking of which, I'm guessing you've checked already, but you can't just compare expectedTermRatingEntity and termRatingQuery?

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,222 @@
...
+    when(request.getParameter("languages")).thenReturn("false");
+    when(request.getParameter("ID")).thenReturn("numberOneId");
+    Entity termEntity = new Entity("Term");
+    Key termKey = termEntity.getKey();
+    when(request.getParameter("Course"))
+        .thenReturn(Course("Spring 2020", "Smith", 12, termKey));
+
+    newTermRating.addTermRating(request);
+
+    Entity termRatingQuery =
```
LukeGMTravis, line 62: This is an entity, not a query (variable name change)

src/test/java/com/google/sps/DataServletTest.java
LukeGMTravis: Mock out the response from the translateService

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -49,65 +60,65 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    // Gets user ID from URL.
+    String userId = request.getParameter("ID");
+    // Gets term key from URL.
+    Key currentTermKey = KeyFactory.stringToKey(request.getParameter("term"));
 
-    response.setContentType("text/html; charset=UTF-8");
-    response.setCharacterEncoding("UTF-8");
-    response.sendRedirect("/index.html");
+    // Check whether user has reviewed that term.
+    List<Entity> termRatingQueryList = queryEntities("Rating", "reviewer-id", userId);
```
LukeGMTravis, line 87: Since the params for `queryEntities` are all strings it's not necessarily clear what they are. One solution to this is to pass an object instead so that it's explicit what the values are. A work around that's often used though is to just comment the individual params when the method is called: ` queryEntities(/* entityName */ "Rating", /* propertyName */ "reviewer-id", /* propertyValue */ userId)`. It would probably be a good idea to add commenting like this to the places where you call `queryEntities`.

src/main/webapp/comments.html
```diff
@@ -0,0 +1,67 @@
...
+    <p>Hello</p>
+    <div id="hours-chart" style="width: 900px; height: 500px"></div>
+    <div id="difficulty-chart" style="width: 900px; height: 500px"></div>
+
+    <p>Rate your term and professors!</p>
+    <form action="/data" method="POST">
+      <p>Rate your term:</p>
+
+      <textarea name="term-input" maxlength="1500">
+Write opinion on term here.</textarea
```
LukeGMTravis, line 23: Formatting might not be so wonky here if the closing textarea tag was on it's own line (like the opening tag above).

src/main/webapp/comments.html
```diff
@@ -0,0 +1,67 @@
...
+      <p>How would you rate this term?</p>
+      <input
+        type="number"
+        id="rating-term"
+        name="rating-term"
+        min="0"
+        max="5"
+      />
+      <label for="rating-term"
+        >Term Rating (From 0 = Terrible to 5 = Excellent)</label
```
LukeGMTravis, line 35: For longer content like this we often put the html tags on their own lines, which would probably also fix the weird prettier formatting.

src/main/webapp/comments.html
```diff
@@ -0,0 +1,67 @@
...
+      >
+
+      <p>How many hours per week did you spend on this term?</p>
+      <input type="number" id="hoursOfWork" name="hoursOfWork" min="1" />
+      <label for="hoursOfWork">Hours (include in-class time)</label>
+
+      <p>How difficult did you find this term?</p>
+      <input type="number" id="difficulty" name="difficulty" min="0" max="5" />
+      <label for="difficulty"
+        >Difficulty (From 0 = Very Easy to 5 = Very Hard)</label
```
LukeGMTravis, line 45: Ditto

src/main/webapp/comments.html
```diff
@@ -0,0 +1,67 @@
...
+      >
+
+      <textarea name="prof-input" maxlength="1500">
+Write opinion on professor here.</textarea
```
LukeGMTravis, line 49: Ditto

src/main/webapp/comments.html
```diff
@@ -0,0 +1,67 @@
...
+      <p>How would you rate the professor?</p>
+      <input
+        type="number"
+        id="rating-prof"
+        name="rating-prof"
+        min="0"
+        max="5"
+      />
+      <label for="rating-prof"
+        >Professor Rating (From 0 = Terrible to 5 = Excellent)</label
```
LukeGMTravis, line 61: Ditto

src/main/webapp/functions.js
```diff
@@ -0,0 +1,90 @@
...
+  fetch("/login")
+    .then((response) => response.json())
+    .then((userDetails) => {
+      const loginElement = document.getElementById("login-box");
+      const schoolName = document.getElementById("school-name");
+      if (!userDetails.loggedIn) {
+        schoolName.style.display = "none";
+        loginElement.innerHTML += `<p><a href= ${userDetails.loginURL} > Click here to log in! </a></p>`;
+      } else {
+        console.log(userDetails);
```
LukeGMTravis, line 57: Left over

src/main/webapp/functions.js
```diff
@@ -0,0 +1,90 @@
...
+        schoolName.style.display = "block";
+        if (userDetails.hasOwnProperty("schoolName")) {
```
LukeGMTravis, line 59: Can just do `if (userDetails.schoolName)` here

src/main/webapp/functions.js
```diff
@@ -0,0 +1,90 @@
...
+/** Contacts Login servlet to verify if user is logged in. */
+function verifyLogin() {
+  fetch("/login")
+    .then((response) => response.json())
+    .then((userDetails) => {
+      const loginElement = document.getElementById("login-box");
+      const schoolName = document.getElementById("school-name");
+      if (!userDetails.loggedIn) {
+        schoolName.style.display = "none";
+        loginElement.innerHTML += `<p><a href= ${userDetails.loginURL} > Click here to log in! </a></p>`;
```
LukeGMTravis, line 55: Nit: No space after `href`

LukeGMTravis: You also technically don't need spacing inside the a tags

src/main/webapp/functions.js
```diff
@@ -0,0 +1,90 @@
...
+      } else {
+        console.log(userDetails);
+        schoolName.style.display = "block";
+        if (userDetails.hasOwnProperty("schoolName")) {
+          schoolName.innerHTML += userDetails.schoolName;
+        }
+        loginElement.innerHTML += `<p><a href= ${userDetails.logoutURL} + > Click here to log out! </a></p>`;
```
LukeGMTravis, line 62: Ditto

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,154 @@
...
+  @AfterEach
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Mock HttpServletRequest request;
+  @Mock LanguageServiceClient languageService;
+
+  @Test
+  public void addTermRating_newRatingNoTranslation() throws IOException {
```
LukeGMTravis, line 55: Can remove the part about "noTranslation" if the translation option has been removed. Same below.

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,154 @@
...
+    request = Mockito.mock(HttpServletRequest.class);
+    when(request.getParameter("term-input")).thenReturn("I do not like this.");
+    when(request.getParameter("rating-term")).thenReturn("1");
+    when(request.getParameter("hoursOfWork")).thenReturn("8");
+    when(request.getParameter("difficulty")).thenReturn("4");
+    when(request.getParameter("prof-input")).thenReturn("The professor was amazing.");
+    when(request.getParameter("rating-professor")).thenReturn("3");
+    when(request.getParameter("ID")).thenReturn("numberOneId");
+
+    Entity termEntity = new Entity("Term");
```
LukeGMTravis, line 67: Suggestion: you can move this block of code up above the block of "whens" above so that the "when" just below this can join the others.

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -0,0 +1,154 @@
...
+    datastore.put(termEntity);
+    Key termKey = termEntity.getKey();
+
+    when(request.getParameter("term")).thenReturn(KeyFactory.keyToString(termKey));
+
+    Document document =
```
LukeGMTravis, line 73: This looks unused. Same in the other test.

src/main/webapp/functions.js
lqueipom: These functions are leftover from Nina's old functions, I will delete them now!

PR #8, opened 2020-07-27
Login

ninaprabhu:
Edit login to use a more formal authorization method and get information like name and generate user ID without using Java servlet.


src/main/webapp/search.html
```diff
@@ -5,10 +5,14 @@
     <title>Search Page</title>
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
+    <script src="https://apis.google.com/js/platform.js" async defer></script>
+    <meta name="google-signin-client_id" content="12523748853-6h3fnh1kppht9s895ist2b5222ifh2u7.apps.googleusercontent.com">
+
   </head>
-  <body onload="loadBody()">
+  <body>
+    <div class="g-signin2" id="login-button" data-onsuccess="onSignIn"></div>
```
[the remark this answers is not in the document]
ninaprabhu: The class is created by Google and is the class that makes the login button do/look like what it does. I don't like the name either :(

src/main/webapp/script.js
```diff
@@ -1,59 +1,15 @@
...
-        };
-        const diffChart = new google.visualization.Histogram(document.getElementById("difficulty-chart"));
-        diffChart.draw(difficultyData, diffOptions);
-    });
-}
-
-function loadBody() {
-  verifyLogin();
+function onSignIn(googleUser) {
+  const profile = googleUser.getBasicProfile();
```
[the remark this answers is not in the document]
ninaprabhu: We went over this in our call earlier, but I'll just write it here in case anyone else is wondering the same thing. onSignIn essentially holds the information that you grant Google when you log in. (Like when you check the box or say "OK" to google asking you for your email, name, etc) and puts that in the profile variable.

src/main/webapp/search.html
```diff
@@ -5,10 +5,14 @@
...
+    <div class="hidden" id="logged-in" >
       <div id="school-name"></div>
-      <div id="login-box"></div>
       <form id="class-form" method="POST" action="/search">
         <p>Enter class info:</p>
         <textarea name="class-name">Class Name</textarea>
```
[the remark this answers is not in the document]
ninaprabhu: Sounds good, I put this in my search PR because it's probably more related to that

src/main/java/com/google/sps/data/AuthenticationServlet.java
```diff
@@ -0,0 +1,68 @@
...
+    userQuery.setFilter(emailFilter);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    List<Entity> results = datastore.prepare(userQuery).asList(FetchOptions.Builder.withDefaults());
+
+    if (status.equals("logged-in")) {
+      String name = request.getParameter("name");
+      String id = request.getParameter("id");
+      if (results.isEmpty()) {
+        Entity newUser = new Entity("User");
+        newUser.setProperty("Name", name);
```
hailed748, line 51: Not really crucial, but I think with all the other entities I was using lowercase for all properties and entity types. I don't think the user entities are accessed outside of this so feel free to keep it the same or change it.

ninaprabhu: Oh shoot I've been using all uppercase, I'll change it to make it easier to combine later

Tetheta:
Starting review

src/main/webapp/loginScript.js
```diff
@@ -0,0 +1,48 @@
...
+  const auth2 = gapi.auth2.getAuthInstance();
+  const profile = auth2.currentUser.get().getBasicProfile();
+  auth2.signOut();
+  const email = profile.getEmail();
+  document.getElementById("class-info").classList.add("hidden");
+  document.getElementById("login-button").classList.remove("hidden");
+  const url = new URL("/authentication", window.location.origin);
+  url.searchParams.set("email", email);
+  url.searchParams.set("status", "logged-out");
+  fetch(url);
```
Tetheta, line 24: Often when we log a user out, we just redirect them to the login page, refreshing everything to ensure there isn't any leftover private information loaded into the page. Might be easier than adding/removing class names. You'll still need to ping the server for the signout though.

src/main/webapp/loginScript.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function showClasses() {
```
Tetheta, line 27: Are we calling things classes or courses? Classes has some potential conflict here with styling classes and the class object in Java/Javascript.

src/main/webapp/loginScript.js
```diff
@@ -0,0 +1,48 @@
+function onSignIn(googleUser) {
+  const profile = googleUser.getBasicProfile();
+  document.getElementById("class-info").classList.remove("hidden");
+  document.getElementById("login-button").classList.add("hidden");
+  document.getElementById("school-name").innerHTML = `Hi, ${profile.getName()}! Your email is ${profile.getEmail()}`;
+  const url = new URL("/authentication", window.location.origin);
+  url.searchParams.set("name", profile.getName()); //TODO: Make this information more secure - not in URL.
```
Tetheta, line 7: Works for now, but good comment about this being a privacy sensitive area.

src/main/java/com/google/sps/data/AuthenticationServlet.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that logs user activity. */
+@WebServlet("/authentication")
+public class AuthenticationServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
Tetheta, line 37: Typically we have one endpoint for login, and another for logout. 
That's a bit awkward with servlet limitations, but just FYI.

src/main/java/com/google/sps/data/AuthenticationServlet.java
```diff
@@ -0,0 +1,68 @@
...
+    String email = request.getParameter("email");
+    String status = request.getParameter("status");
+    Filter emailFilter = new FilterPredicate("email", FilterOperator.EQUAL, email);
+    Query userQuery = new Query("User");
+    userQuery.setFilter(emailFilter);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    List<Entity> results = datastore.prepare(userQuery).asList(FetchOptions.Builder.withDefaults());
+
+    if (status.equals("logged-in")) {
```
Tetheta, line 46: I imagine you know this, but just checking a param like this to determine if someone's logged in is 100% insecure. 
With this current implementation I could open a new tab and sign in as any user as long as I know their email.

For now that's probably fine, at least we have a semblance of login flow going on here.

Just an FYI that you've done one step (the front end implementation) of a very complex process. Properly storing user data on the server can get enormously complicated.

src/main/webapp/search.html
```diff
@@ -4,11 +4,15 @@
     <meta charset="UTF-8">
     <title>Search Page</title>
     <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
+    <script src="loginScript.js"></script>
```
Tetheta, line 7: With the simple site you're creating here, it might be easier to have a dedicated login.html, then a logout button that redirects to that page after signaling the server to log out.

Then if a user goes to a page but isn't logged in, it will redirect them to that page first.

Otherwise you'll have to put this on every page, and handle some weird edge cases where they click between pages but the server response hasn't come back yet/etc causing user frustration as they have to login yet again.

Tetheta: As with most of my comments here, this could be done in a future PR, but it's something to keep in mind.

src/main/webapp/search.html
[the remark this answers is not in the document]
Tetheta: Yeah not impressed with their demo example :( fine for now.

src/main/webapp/search.html
```diff
@@ -4,11 +4,15 @@
...
+    <script src="https://apis.google.com/js/platform.js" async defer></script>
+    <meta name="google-signin-client_id" content="12523748853-6h3fnh1kppht9s895ist2b5222ifh2u7.apps.googleusercontent.com">
+
   </head>
-  <body onload="loadBody()">
+  <body>
+    <div class="g-signin2" id="login-button" data-onsuccess="onSignIn"></div>
+    <div class="hidden" id="class-info" >
```
Tetheta, line 14: FYI it's trivially easy to unhide this in production if the user knows anything about html. So you'll need to make sure you guard the search query on the server side against not-logged-in requests. 

Granted that's something you should be doing anyways for all server endpoints.

src/main/webapp/loginScript.js
ninaprabhu: Hmmm I changed this to showCourses() - not sure why it's not showing up in this PR

ninaprabhu: OH I actually did all of the class -> course in search

PR #11, opened 2020-07-31
Adds ability to find term based on URL parameters

hailed748:
Starting use of live data

Adds: New servlet that simply takes in URL parameters and finds the term entity the rating data needs to be pulled from, I also added one test cases to make sure this was being done properly. Will add more test for invalid URL's

### DIFF
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -24,7 +24,6 @@ public class LiveCourseData extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    findData(db, request);
     response.setContentType("text/html; charset=UTF-8");
     response.setCharacterEncoding("UTF-8");
     response.sendRedirect("/AddSchoolData.html");
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -1,0 +1,78 @@
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.CompositeFilterOperator;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -0,0 +1,78 @@
...
+    String courseName = request.getParameter("course-name");
+    String termName = request.getParameter("term");
+    String profName = request.getParameter("professor-name");
+    Long units = Long.parseLong(request.getParameter("units"));
+
+    Entity foundTerm = findTerm(db, schoolName, courseName, termName, units, profName);
+    return foundTerm;
+  }
+
+  private Entity findTerm(
```
ninaprabhu, line 44: why is one of these public and the other private?

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,108 @@
...
+    Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    return result;
+  }
+
+  private List<Entity> findQueryMatch(DatastoreService db, String entityType) {
+    Query q = new Query(entityType);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    return result;
+  }
```
ninaprabhu, line 107: Is the second function necessary? Or is it  a duplicate of the first

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: It needs to be public so the test file can find it.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
hailed748: Not necessary I was using it to debug something.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,102 @@
...
+  public void FindingExisitngTermEntity() {
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    request = createRequest(request, "MIT", "6.006", "Spring 2020", "12", "Jason Ku");
+    requestB = createRequest(requestB, "MIT", "6.008", "Spring 2018", "6", "Srini");
+    String expectedTermName = "Spring 2020";
+
+    newSchoolObject.addSchoolData(db, request);
+    newSchoolObject.addSchoolData(db, requestB);
+    Key expectedParent = findQueryMatch(db, "Course", "course-name", "6.006").get(0).getKey();
+    Entity found = LiveData.getTerm(db, request);
```
ninaprabhu, line 73: IDK if this also applies to you but apparently we're supposed to be testing the whole doGet function here?

hailed748: I tried making it doGet but it won't work because it expects doGet to be void, but in order to test I have to return something.

Tetheta:
Starting review

Tetheta:
This looks like a dependent PR on the other one I just reviewed. It's hard to tell what should've come first.

One way to make that more clear is that instead of creating a PR against master, create a PR to merge into your other branch. That way the diff contains only the changed code, and when the other branch gets merged, this one is simple to update and merge as well. 

Right now it's difficult for me to tell what's different from the other PR. They look basically identical but I think the other has more so all my comments there apply.

hailed748:
I am almost certain the other branch branches off of this one, not really sure what went wrong.

PR #12, opened 2020-08-01
Gathers Ratings

hailed748:
Adds function that collects hours/difficulty ratings.

### DIFF
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -24,13 +24,13 @@ public class LiveCourseData extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    findData(db, request);
+    findTerm(db, request);
     response.setContentType("text/html; charset=UTF-8");
     response.setCharacterEncoding("UTF-8");
     response.sendRedirect("/AddSchoolData.html");
   }
 
-  public Entity findData(DatastoreService db, HttpServletRequest request) {
+  public Entity findTerm(DatastoreService db, HttpServletRequest request) {
     String schoolName = request.getParameter("school-name");
     String courseName = request.getParameter("course-name");
     String termName = request.getParameter("term");
@@ -39,6 +39,16 @@ public class LiveCourseData extends HttpServlet {
 
     Entity foundTerm = findTerm(db, schoolName, courseName, termName, units, profName);
     return foundTerm;
+  }
+
+  public List<Long> getTermData(DatastoreService db, Entity term, String property) {
+    List<Long> dataList = new ArrayList();
+    List<Entity> termRatings = findChildren(db, "Rating", term.getKey());
+
+    for (Entity rating : termRatings) {
+      dataList.add(Long.parseLong(rating.getProperty(property).toString()));
+    }
+    return dataList;
   }
 
   private Entity findTerm(
@@ -72,6 +82,12 @@ public class LiveCourseData extends HttpServlet {
       DatastoreService db, String entityType, String entityProperty, String propertyValue) {
     Filter filter = new FilterPredicate(entityProperty, FilterOperator.EQUAL, propertyValue);
     Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    return result;
+  }
+
+  private List<Entity> findChildren(DatastoreService db, String type, Key parent) {
+    Query q = new Query(type).setAncestor(parent);
     List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
     return result;
   }
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -41,7 +41,7 @@ public class LiveCourseData extends HttpServlet {
     return foundTerm;
   }
 
-  public List<Long> getTermData(DatastoreService db, Entity term, String property) {
+  public List<Long> getDataFromTermRating(DatastoreService db, Entity term, String property) {
     List<Long> dataList = new ArrayList();
     List<Entity> termRatings = findChildren(db, "Rating", term.getKey());
 
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -46,7 +46,7 @@ public class LiveCourseData extends HttpServlet {
     List<Entity> termRatings = findChildren(db, "Rating", term.getKey());
 
     for (Entity rating : termRatings) {
-      dataList.add(Long.parseLong(rating.getProperty(property).toString()));
+      dataList.add((long) rating.getProperty(property));
     }
     return dataList;
   }
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -24,13 +24,13 @@ public class LiveCourseData extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    findTerm(db, request);
+    getTerm(db, request);
     response.setContentType("text/html; charset=UTF-8");
     response.setCharacterEncoding("UTF-8");
     response.sendRedirect("/AddSchoolData.html");
   }
 
-  public Entity findTerm(DatastoreService db, HttpServletRequest request) {
+  public Entity getTerm(DatastoreService db, HttpServletRequest request) {
     String schoolName = request.getParameter("school-name");
     String courseName = request.getParameter("course-name");
     String termName = request.getParameter("term");
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -87,8 +87,8 @@ public class LiveCourseData extends HttpServlet {
   }
 
   private List<Entity> findChildren(DatastoreService db, String type, Key parent) {
-    Query q = new Query(type).setAncestor(parent);
-    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    Query children = new Query(type).setAncestor(parent);
+    List<Entity> result = db.prepare(children).asList(FetchOptions.Builder.withDefaults());
     return result;
   }
 }
--- a/src/main/java/com/google/sps/servlets/LiveCourseData.java
+++ b/src/main/java/com/google/sps/servlets/LiveCourseData.java
@@ -1,0 +1,78 @@
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.CompositeFilterOperator;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -0,0 +1,94 @@
...
+    filters.add(courseFilter);
+    filters.add(unitFilter);
+
+    Query courseQuery =
+        new Query("Course").setAncestor(schoolKey).setFilter(CompositeFilterOperator.and(filters));
+    Key courseKey =
+        db.prepare(courseQuery).asList(FetchOptions.Builder.withDefaults()).get(0).getKey();
+
+    Filter termFilter = new FilterPredicate("term", FilterOperator.EQUAL, termName);
+    Query termQuery = new Query("Term").setAncestor(courseKey).setFilter(termFilter);
```
lqueipom, line 75: Could this query not return anything? Then, wouldn't findTerm(db, request) might return nothing?

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -0,0 +1,94 @@
...
+  private List<Entity> findQueryMatch(
+      DatastoreService db, String entityType, String entityProperty, String propertyValue) {
+    Filter filter = new FilterPredicate(entityProperty, FilterOperator.EQUAL, propertyValue);
+    Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    return result;
+  }
+
+  private List<Entity> findChildren(DatastoreService db, String type, Key parent) {
+    Query q = new Query(type).setAncestor(parent);
```
lqueipom, line 90: Maybe use a more descriptive variable name for the query.

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: I don't think it will ever be empty because this request is coming from a URL that Nina will make when a course is found. That means that it must exists in our DB.

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -0,0 +1,94 @@
...
+    Entity foundTerm = findTerm(db, schoolName, courseName, termName, units, profName);
+    return foundTerm;
+  }
+
+  public List<Long> getTermData(DatastoreService db, Entity term, String property) {
+    List<Long> dataList = new ArrayList();
+    List<Entity> termRatings = findChildren(db, "Rating", term.getKey());
+
+    for (Entity rating : termRatings) {
+      dataList.add((long) rating.getProperty(property));
```
ninaprabhu, line 49: Would long be capitalized here?

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Mock HttpServletRequest request;
+  @Mock HttpServletRequest requestB;
+
+  @Test
+  public void FindingExisitngTermEntity() {
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    request = createRequest(request, "MIT", "6.006", "Spring 2020", "12", "Jason Ku");
+    requestB = createRequest(requestB, "MIT", "6.008", "Spring 2018", "6", "Srini");
```
ninaprabhu, line 69: Luke commented this on my PR - maybe put comments in the request parameters to specify what variable each is? Like createRequest(request, /*schoolname*/ "MIT", ...)

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: it actually needs to be lowercase when you cast, upper leads to an error

Tetheta:
Starting review

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    getTerm(db, request);
+    response.setContentType("text/html; charset=UTF-8");
+    response.setCharacterEncoding("UTF-8");
+    response.sendRedirect("/AddSchoolData.html");
+  }
+
+  public Entity getTerm(DatastoreService db, HttpServletRequest request) {
```
Tetheta, line 33: I can understand this needing to be public for testing (but if that's the case please leave a comment as much).

However the rest should be private unless I'm mistaken.

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -0,0 +1,94 @@
...
+    String courseName = request.getParameter("course-name");
+    String termName = request.getParameter("term");
+    String profName = request.getParameter("professor-name");
+    Long units = Long.parseLong(request.getParameter("units"));
+
+    Entity foundTerm = findTerm(db, schoolName, courseName, termName, units, profName);
+    return foundTerm;
+  }
+
+  public List<Long> getTermData(DatastoreService db, Entity term, String property) {
```
Tetheta, line 44: Data is not a helpful name usually. It looks like this is just ratings for now so maybe call it that? Easy enough to change later.

When you do change, then ideally we'd create a record of some sort to denote what kind of data we're passing around.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private HttpServletRequest createRequest(
+      HttpServletRequest request,
+      String schoolName,
+      String courseName,
+      String termName,
+      String units,
+      String profName) {
+    request = Mockito.mock(HttpServletRequest.class);
```
Tetheta, line 112: If you're already declaring a new on here, why pass it in as a parameter? I don't think that's necessary. 

Also both returning a type while also asking for the same type in the params confuses people. It's not clear then if mutation is going on or what exactly.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+public final class LiveCourseDataTest {
+  private static final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+
+  private AddSchoolData newSchoolObject;
+  private LiveCourseData LiveData;
```
Tetheta, line 47: Be consistent please. If it's a class, it's UpperCamelCase, but if it's an instance of a class, it's lowerCamelCase. 

Convention for these things is often to use the class name, just convert it to lowerCamelCase (so here we'd have addSchoolData and liveCourseData as the variable names).

Tetheta: Though it looks like you have a method addSchoolData on AddSchoolData so maybe don't rename it so we can all keep our sanity. That's another issue we probably should've caught but not a big deal.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    assertEquals(expectedParent, found.getParent());
+    assertEquals(expectedTermName, found.getProperty("term"));
+  }
+
+  @Test
+  public void GettingRatingData_Hours_Difficulty() {
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    Entity parent = addTermEntity(db);
+    addRatingEntity(db, 12, 7, parent.getKey());
```
Tetheta, line 92: Annotating these numbers with what they're referring to can be helpful: `/* hours */ 12, /* difficulty */ 7`

src/main/java/com/google/sps/servlets/LiveCourseData.java
Tetheta: Ah I see from the test you're getting metadata from the ratings, not the ratings themselves. 
In that case, something like getTermDataFromRatngs or something could be helpful. Up to you.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
hailed748: Nice catch, I was even confused as to why I did that.

hailed748: I actually did a little more testing. And what I am doing now is passing in the @Mock request and @Mock requestB I define at the top. I then reassign these inside of the function. When I try doing it the simpler way of just assigning request = createRequest(...) and requestB = createRequest(...) I get an IndexOutOfBounds error. Not sure why this is the case.

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: I actually call getTermData in my tests as well so it needs to be public. I will refactor my code soon so that I only need to have on public function for testing.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
Tetheta: Huh, weird. Well as long as you're consistent it's fine.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -89,14 +89,15 @@ public void FindingExisitngTermEntity() {
   public void GettingRatingData_Hours_Difficulty() {
     DatastoreService db = DatastoreServiceFactory.getDatastoreService();
     Entity parent = addTermEntity(db);
-    addRatingEntity(db, 12, 7, parent.getKey());
+    addRatingEntity(db, /* hours */ 12, /* difficulty */ 7, parent.getKey());
     addRatingEntity(db, 7, 4, parent.getKey());
```
Tetheta, line 93: Nit: would be nice to add the hours/difficulty annotations here too

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
Tetheta: I see you've requested a re-review but I don't see the changes mentioned here yet?

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
hailed748: Just wondering for the future. If I have a case where I am adding like 5 entities would i need to add a comment for each one?

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
hailed748: Wait is there an error here outside of me passing in the request?

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
Tetheta: It should be just copy/paste yes? We only do it when it's not obvious what you're passing (so raw primitives basically, ints, longs, strings).

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
Tetheta: My understanding from your comment is that you were going to change this to just mutate the request (which is what you are already doing).

However that implies you'd change where this is used to not expect a return object (and also change the return type of this function to void). 

I'll comment where it's use to further explain.

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -0,0 +1,143 @@
...
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    request =
+        createRequest(
+            request, /*schoolName*/
+            "MIT", /*courseName*/
+            "6.006", /*termName*/
+            "Spring 2020", /*units*/
+            "12", /*profName*/
+            "Jason Ku");
+    requestB = createRequest(requestB, "MIT", "6.008", "Spring 2018", "6", "Srini");
```
Tetheta, line 76: Up here, this is really confusing to look at. You're both passing in something like you expect it to be mutated, but you're also assigning to that same variable. 

As someone reading this I now have no idea which I should be relying on and have to dive into the depths of your function to figure it out.

Choose one. I believe from the comment earlier that mutating works better, so remove the return type and don't reassign here.

PR #14, opened 2020-08-04
Course page boiler

hailed748:
Adds boiler plate for course page. Still missing some sections, just wanted to get something started. Also puts back some of the hard-coded data for the time being.

image (3)


src/main/webapp/course-style.css
```diff
@@ -0,0 +1,80 @@
...
+}
+
+.flex-child {
+    flex: 1;
+}  
+
+.form-container {
+    padding-top: 100px;
+}
+
```
lqueipom, line 13: Nit: two extra lines

src/main/webapp/course-style.css
```diff
@@ -0,0 +1,80 @@
...
+  display: inline-block;
+  font-size: larger;
+  margin: 8px 30px;
+  padding: 12px 20px;
+}
+
+#form-container {
+  background-color: #f2f2f2;
+  border-radius: 5px;
+  padding: 100px 0;
```
lqueipom, line 58: Nit: alphabetical order

Tetheta:
Starting review

src/main/webapp/course-style.css
```diff
@@ -0,0 +1,78 @@
...
+  padding: 12px 20px;
+  text-align: center;
+  width: 70%;
+}
+
+input[type="submit"] {
+  background-color: #4caf50;
+  border: none;
+  border-radius: 4px;
+  color: white;
```
Tetheta, line 35: FYI colors like this ("white", "black", etc) often are different across browsers. So in prod code we always specify in hex instead.

src/main/webapp/course-style.css
```diff
@@ -0,0 +1,78 @@
...
+
+.my-3 {
+    margin: 20px 125px;
+}
+
+input[type="number"],
+textarea {
+  border: 1px solid #228b22;
+  border-radius: 4px;
+  box-sizing: border-box;
```
Tetheta, line 22: border-box is often something you want for either everything or nothing so might be worth putting at the top level. 

Often people like to do something like
```
/* Remove padding and margin */ 
*  
    { 
    margin: 0; 
    padding: 0; 
    border: 0; 
}
```

In some index.css that gets added to every page (* means it hits all elements). 
Adding box-sizing to something like this could be good.

Btw * as a selector isn't performant technically but it hardly matters for your use case.

src/main/webapp/course-style.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#form-container {
+  align-items: center;
+  background-color: #f2f2f2;
+  border-radius: 5px;
+  display:flex;
+  flex-flow: column wrap;
+  justify-content: center;
+  padding: 100px 0;}
```
Tetheta, line 61: Funky ending bracket here.

PR #15, opened 2020-08-05
Styling/connecting with chart functionality

ninaprabhu:
image

Implemented some styling (planning on improving) and changed filenames of .js and .css files to not conflict with anyone else's branches


Tetheta:
Starting review. Normally would wait for Laura but this is time-sensitive.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -127,15 +130,21 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     Long units = Long.parseLong(request.getParameter("num-units"));
-    String term = request.getParameter("term-name");
+    String term = request.getParameter("term");
+    String school = request.getParameter("school-name");
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Entity newCourse = new Entity("Course");
+
+    AddSchoolData addSchool = new AddSchoolData();
+    addSchool.addSchoolData(datastore, request);
```
Tetheta, line 139: This might give odd behavior, not sure what instantiating a servlet inside a servlet will do.

Keep that in mind if you have issues, might want to extract the majority of the functionality to a data class. 

Fine for the demo though.

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,84 @@
+function countSearchUnits() {
```
Tetheta, line 1: If you're populating an array, then you're not counting right? 
Renaming to something like collectUnitsFromSearch might make more sense.

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,84 @@
...
+  const arr = []
+  for (let i = 0; i < inputElems.length; i++) {       
+    if (inputElems[i].checked){
+      arr.push(inputElems[i].value);
+    }
+  }
+  return arr;
+}
+
+function countAddUnits() {
```
Tetheta, line 12: Same comment

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,84 @@
...
+  const inputElems = document.getElementsByName("num-units");
+  const arr = []
```
Tetheta, line 14: arr is never a good variable name by the way. 
Something like addedUnits if that's what you're adding here might be better.

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,84 @@
...
+  for (let i = 0; i < inputElems.length; i++) {
```
Tetheta, line 15: FYI in javascript you can use for of:
```
for(let elem of inputElems) {
 // Code
}
```

Usually cleaner.

Or in this case, this is a great case for doing an array filter then a map operation that would turn this into basically a one liner. That can wait for later though.

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,84 @@
...
+  const termName = document.getElementById("search-term").value;
+  const units = countSearchUnits();
+  const school = getUserSchool();
+  courseResults.innerHTML = "";
+  const url = new URL("/search", window.location.origin);
+  url.searchParams.set("courseName", courseName);
+  url.searchParams.set("profName", profName);
+  url.searchParams.set("units", units);
+  url.searchParams.set("term", termName);
+  url.searchParams.set("schoolName", getUserSchool())
```
Tetheta, line 36: Why inline this call but not the others? Be consistent in this please. Either way works

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,84 @@
...
+  fetch(url)
+  .then(response => response.json())
+  .then((courses) => {
+    courses.forEach(course => courseResults.appendChild(createListElement(course)));
+  });
```
Tetheta, line 41: Eventually we'll want to add error handling, but that can wait for the polishing stage.

src/main/webapp/search-style.css
```diff
@@ -0,0 +1,42 @@
...
+}
+
+label, select {
+  margin-bottom: 20px;
+}
+
+#login-box {
+    margin:auto;
+    width: 120px;
+    padding: 30px;
```
Tetheta, line 12: Weird extra spacing here. Is prettier doing this?

src/main/webapp/search-style.css
```diff
@@ -0,0 +1,42 @@
+.hidden {
```
Tetheta, line 1: The whole file is a bit wonky in terms of formatting (not alphabetical either). Is Prettier not running on css files or was it causing issues?
Not something for this PR but something I'd like to know about for future use in stuff like this.

src/main/webapp/search-style.css
```diff
@@ -0,0 +1,42 @@
...
+    margin:auto;
+    width: 120px;
+    padding: 30px;
+}
+
+#search-textarea, #search-selectboxes, #class-textarea, #class-selectboxes {
+    display: flex;
+}
+
+textarea, select, fieldset {
```
Tetheta, line 19: Keep in mind that this will affect everything on the page (and thus can easily conflict). Often we prefer to put styles like this in the top level index.css file or just don't use them at all.

I'll let it slide for now.

src/test/java/com/google/sps/servlets/DataServletTest.java
```diff
@@ -0,0 +1,144 @@
+// package com.google.sps.servlets;
+
```
Tetheta, line 2: Why add a giant commented file?

ninaprabhu: Sorry - didn't mean to actually commit this. I commented out this test file because there were bugs that whoever owned it hadn't committed + fixed yet.

src/main/webapp/search-style.css
ninaprabhu: I think so - I'll go through and try to fix it. Prettier is being pretty weird.

src/main/webapp/search-style.css
ninaprabhu: I'll go through and fix the alphabetical order and see what I can do about the spacing. Sorry about that - I think it's a combination of me being in a rush and Prettier being weird.

src/main/webapp/search-style.css
ninaprabhu: Since I'm only calling this in search.html, I figured it would only affect those things on search.html (which is what I intended) - or is that still bad style?

src/main/webapp/search-script.js
ninaprabhu: oh that's awesome, didn't know you could do that, will implement right now (and later I'll do the map operation, definitely seems nicer)

src/main/webapp/search-script.js
ninaprabhu: Oops - just noticed that I set const school above but forgot to actually call the school variable in this line. Changed!

src/main/webapp/search-style.css
Tetheta: It's still risky if you end up adding other css files to search.html but much better than I was thinking. I'm too used to single page apps haha. Don't worry about changing it.

PR #16, opened 2020-08-05
Styling form

lqueipom:


lqueipom:
Screenshot 2020-08-05 at 8 01 11 PM

Improved form styling, I am attaching here a picture.

pom.xml
```diff
@@ -61,12 +61,6 @@
         <version>5.4.2</version>
         <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <version>5.4.2</version>
-      <scope>test</scope>
-    </dependency>
```
Tetheta, line 69: There a particular reason you're deleting these?

src/main/webapp/comments.html
```diff
@@ -3,65 +3,90 @@
   <head>
     <meta charset="UTF-8" />
     <title>Capstone</title>
-    <link rel="stylesheet" href="style.css" />
+    <link rel="stylesheet" href="styleComments.css" />
```
Tetheta, line 6: You all should probably get together and decide if you're using lowerCamelCase or dash-case for names like this. Just be consistent, I think Nina's using dash-case.

src/main/webapp/styleComments.css
```diff
@@ -0,0 +1,58 @@
+input[type="number"],
```
Tetheta, line 1: Be aware that general tags like this will start to conflict if anyone adds another stylesheet to your page. Fine for now though.

src/test/java/com/google/sps/DataServletTest.java
```diff
@@ -1,142 +1,144 @@
...
-    assertEquals(Long.valueOf(1), termRatingEntity.getProperty("perception-term"));
-    assertEquals(Long.valueOf(10), termRatingEntity.getProperty("hours"));
-    assertEquals(Long.valueOf(5), termRatingEntity.getProperty("difficulty"));
-    assertEquals("This teacher was wonderful.", termRatingEntity.getProperty("comments-professor"));
-    assertEquals(Long.valueOf(3), termRatingEntity.getProperty("perception-professor"));
-    assertEquals(-0.699999988079071, termRatingEntity.getProperty("score-term"));
-    assertEquals(-0.699999988079071, termRatingEntity.getProperty("score-professor"));
-  }
-}
+// package com.google.sps.servlets;
```
Tetheta, line 1: Please either delete or leave it alone. Don't commit commented code. Version control has a copy of it if you need it later.

pom.xml
lqueipom: Yes! There were in my pom file twice for some reason.

src/main/webapp/comments.html
lqueipom: I will use dash case then :)

src/test/java/com/google/sps/DataServletTest.java
lqueipom: File is supposed to be there but didn't want to run test cases, will un-comment now!

src/main/webapp/comments.html
lqueipom: Saw that Dagmawi is also using camel so I think we will get together and figure that out soon.

PR #17, opened 2020-08-05
Live Data

hailed748:
This now pull all the data from an entity. Also changed some variable names to make it consistent so my code can connect to Nina's and Laura's.


src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -84,21 +85,38 @@ public void FindingExisitngTermEntity() {
...
-
-    assertEquals(expectedHoursList, actualHoursList);
-    assertEquals(expectedDifficultyList, actualDifficultyList);
+    createRequest(
+        request, /*schoolName*/
+        "MIT", /*courseName*/
+        "6.006", /*termName*/
+        "Spring 2020", /*units*/
+        "12", /*profName*/
+        "Jason Ku");
```
ninaprabhu, line 96: Nit: adjust this so the name and description comment are on the same line

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -24,29 +25,46 @@
...
+    termDataHolder.setProfessorPerceptionList(
+        getDataFromTermRating(db, foundTerm, "perception-professor"));
+    termDataHolder.setProfessorScoreList(getDataFromTermRating(db, foundTerm, "score-professor"));
+    termDataHolder.setTermCommentsList(getDataFromTermRating(db, foundTerm, "comments-term"));
+    termDataHolder.setProfessorCommentsList(
+        getDataFromTermRating(db, foundTerm, "comments-professor"));
+    return termDataHolder;
+  }
+
   public Entity getTerm(DatastoreService db, HttpServletRequest request) {
```
LukeGMTravis, line 51: Looks like this is no longer used?

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -24,29 +25,46 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    getTerm(db, request);
+    getAllDataFromTerm(db, request);
```
LukeGMTravis, line 28: Do you mean to use response from `getAllDataFromTerm`?

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -84,21 +85,38 @@ public void FindingExisitngTermEntity() {
...
-    List<Long> expectedDifficultyList = new ArrayList(Arrays.asList((long) 7, (long) 4));
-
-    List<Long> actualHoursList = liveCourseData.getDataFromTermRating(db, parent, "hours");
-    List<Long> actualDifficultyList =
-        liveCourseData.getDataFromTermRating(db, parent, "difficulty");
-
-    assertEquals(expectedHoursList, actualHoursList);
-    assertEquals(expectedDifficultyList, actualDifficultyList);
+    createRequest(
+        request, /*schoolName*/
```
LukeGMTravis, line 91: Nit: Comments should be in the form `/* ... */`, that way prettier will put them on the right line

src/test/java/com/google/sps/servlets/LiveCourseDataTest.java
```diff
@@ -123,16 +141,31 @@ private void createRequest(
     return result;
   }
 
-  private void addRatingEntity(DatastoreService db, int hours, int difficulty, Key parent) {
+  private void addRatingEntity(
```
LukeGMTravis, line 144: Since this method has so many params and they are often the same type it's good practice to comment what the params are when the method is called (you can see other examples of this above).

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: I actually use thus function at the top of my getAllDataFromTerm() function

LukeGMTravis: Huh no idea how I missed that lol

PR #18, opened 2020-08-06
Fixing how I get term key

lqueipom:
Modified DataServlet so I get the term key from the school name, course name, units, professor name, and term name (as this is now being passed into the URL instead of the string term key).

Added the getTerm function to DataServlet.

Modified how I pass data to the servlet.


src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+  const units = urlParams.get("num-units");
+  const schoolName = urlParams.get("school-name");
+
+  data["course-name"] = courseName;
+  data["term"] = term;
+  data["prof-name"] = profName;
+  data["units"] = units;
+  data["school-name"] = schoolName;
+
+  const termInput = document.getElementsByName("term-input")[0].value;
```
hailed748, line 152: Maybe change these to getElementById I think I saw that you add IDs to them in the HTML

Tetheta:
Starting review

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -16,28 +16,28 @@
...
 /** An item on a todo list. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
   private final List<Object> commentsList = new ArrayList<>();
   private Key currentTermKey;
-  private final LanguageServiceClient languageService;
+  private LanguageServiceClient languageService;
+  private final DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+  private String schoolName, courseName, profName, termName, termFeedback, professorFeedback;
+  private Long units, termRating, professorRating, workHours, difficulty;
```
Tetheta, line 38: Do these all need to be global?

Typically best practice is to declare variables as close to where they're used as possible. 

Back in the day declaring at the top was useful for memory estimation/etc, but that's not a concern nowadays and it easy to forget to remove one from being declared if it's not co-located.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -63,36 +63,53 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
   public void addTermRating(HttpServletRequest request) throws IOException {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    String termFeedback = request.getParameter("term-input");
-    Long termRating = Long.parseLong(request.getParameter("rating-term"));
-    Long workHours = Long.parseLong(request.getParameter("hoursOfWork"));
-    Long difficulty = Long.parseLong(request.getParameter("difficulty"));
-    String professorFeedback = request.getParameter("prof-input");
-    Long professorRating = Long.parseLong(request.getParameter("rating-professor"));
+
+    StringBuilder jb = new StringBuilder();
```
Tetheta, line 71: why not just call this stringBuilder? jb means nothing to me.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -63,36 +63,53 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    }
 
     float termScore = getSentimentScore(termFeedback);
     float professorScore = getSentimentScore(professorFeedback);
 
-    // Gets user ID from URL.
-    String userId = request.getParameter("ID");
-    // Gets term key from URL.
-    Key currentTermKey = KeyFactory.stringToKey(request.getParameter("term"));
+    // Quick change, will modify tests as well after the demo.
```
Tetheta, line 101: It's post demo now. Is this still the change we want to submit?

src/main/webapp/course.html
```diff
@@ -18,54 +26,67 @@
...
+          Donec id elit non mi porta gravida at eget metus. Fusce dapibus,
+          tellus ac cursus commodo, tortor mauris condimentum nibh, ut fermentum
+          massa justo sit amet risus.
+        </p>
+      </div>
     </div>
 
     <div id="form-container">
-      <form id="term-form">
+      <form id="termForm">
```
Tetheta, line 89: Why change this to camelCase but keep the rest dash-case? Typically html properties are dash-case. I know it's weird that different things are different cases. However once you get used to it it lets you quickly identify when the context of the code you're referencing changed.

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+
+async function postData(url, data = {}) {
+  // Default options are marked with *
+  const response = await fetch(url, {
+    method: "POST", // *GET, POST, PUT, DELETE, etc.
+    cache: "no-cache",
+    credentials: "same-origin", // include, *same-origin, omit
+    headers: {
+      "Content-Type": "application/json",
+      // 'Content-Type': 'application/x-www-form-urlencoded',
```
Tetheta, line 129: Remove this line if it's not needed

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+    },
+  };
+  const profPerceptionChart = new google.visualization.Histogram(
+    document.getElementById("prof-chart")
+  );
+  profPerceptionChart.draw(profPerceptionData, profPerceptionOptions);
+}
+
+async function postData(url, data = {}) {
+  // Default options are marked with *
```
Tetheta, line 122: Super nit: 
Comments here should start with a capital and end with a period (I know this is likely copy pasted, but still good practice to update them).

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+    method: "POST", // *GET, POST, PUT, DELETE, etc.
+    cache: "no-cache",
+    credentials: "same-origin", // include, *same-origin, omit
+    headers: {
+      "Content-Type": "application/json",
+      // 'Content-Type': 'application/x-www-form-urlencoded',
+    },
+    body: JSON.stringify(data), // body data type must match "Content-Type" header
+  });
+  return response.json(); // parses JSON response into native JavaScript objects
```
Tetheta, line 133: Could also note that this returns a promise.

Internally at Google we have a whole type system for that. By default JavaScript doesn't (though TypeScript does).

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+}
+
+function storeData() {
+  var data = {};
```
Tetheta, line 137: no vars please, use let here

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+  const hours = document.getElementById("hoursOfWork").value;
+  const diff = document.getElementById("difficulty").value;
+  document.getElementById("termForm").reset();
+
+  data["term-input"] = termInput;
+  data["prof-input"] = profInput;
+  data["rating-term"] = ratingTerm;
+  data["rating-prof"] = ratingProf;
+  data["hours"] = hours;
+  data["difficulty"] = diff;
```
Tetheta, line 165: I'd use the shorthand object creation here instead.

So this would turn into:
```
const data = {
   "course-name": urlParams.get("course-name"),
   ...
   "term-input": document.getElementById("term-input").value,
}
```

Little easier to read

If you use camelCase, then when constructing the new url pass data.schoolName etc instead.

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+  var data = {};
+  const queryString = window.location.search;
+  const urlParams = new URLSearchParams(queryString);
+  const courseName = urlParams.get("course-name");
+  const term = urlParams.get("term");
+  const profName = urlParams.get("prof-name");
+  const units = urlParams.get("num-units");
+  const schoolName = urlParams.get("school-name");
+
+  data["course-name"] = courseName;
```
Tetheta, line 146: Is it really particular that it needs to be dash-case here? Typically when storing something in JSON we use camelCase.

src/main/webapp/courseLoader.js
Tetheta: Also, you don't need the "" if you're not using dash-case.

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+  term,
+  termInput,
+  profInput,
+  ratingTerm,
+  ratingProf,
+  hours,
+  difficulty
+) {
+  const url = new URL("/data", window.location.origin);
+  url.searchParams.set("course-name", courseName);
```
Tetheta, line 198: If you do keep the data object as dash case, you could pass it in directly here.

Then do this:

```
for(let key in data) {
  url.searchParams.set(key, data[key]);
}
```
Assuming you want all the properties on data to be search params.

src/main/webapp/courseLoader.js
Tetheta: Also data doesn't tell me a whole lot. What sort of data is this? Would be useful to rename these functions to be more explicit about that.

src/main/webapp/courseLoader.js
```diff
@@ -2,126 +2,216 @@ google.charts.load("current", { packages: ["corechart"] });
...
+  url.searchParams.set("term-input", termInput);
+  url.searchParams.set("prof-input", profInput);
+  url.searchParams.set("rating-term", ratingTerm);
+  url.searchParams.set("rating-prof", ratingProf);
+
+  return url;
+}
+
+function postDataForm() {
+  const urlAndData = storeData();
```
Tetheta, line 215: The name "storeData" doesn't really imply that it returns anything. Perhaps getDataToStore? Or better yet, getSearchDataToStore or something if that's what you're actually doing

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -63,36 +63,53 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    String professorFeedback = request.getParameter("prof-input");
-    Long professorRating = Long.parseLong(request.getParameter("rating-professor"));
+
+    StringBuilder jb = new StringBuilder();
+    String line = null;
+    try {
+      BufferedReader reader = request.getReader();
+      while ((line = reader.readLine()) != null) jb.append(line);
+    } catch (Exception exception) {
+      /*report an error*/
```
Tetheta, line 77: Right now this will silently fail, do you want to throw an error or something here?

Tetheta: FYI typically an empty catch statement is an anti-pattern, as it hides errors from you at runtime. Ideally either let it fail in some way (throwing or not catching), or tell in a comment exactly why this isn't actually an error and it's fine to silently catch it.

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: They don't need to be global, I will fix that now!

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: Yes! This is still the change we want to submit, I am re-formatting my tests cases on a separate PR that I am working on now.

src/main/webapp/course.html
lqueipom: Ohhh that's interesting! Will keep that in mind :)

src/main/webapp/courseLoader.js
lqueipom: I am changing it to dash case and I later want to add the user ID to data and do not want that to be on the URL.

lqueipom: I am changing it from*. It will now be camel case.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -67,27 +65,38 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
   public void addTermRating(HttpServletRequest request) throws IOException {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-
-    StringBuilder jb = new StringBuilder();
+    StringBuilder stringBuilder = new StringBuilder();
     String line = null;
     try {
       BufferedReader reader = request.getReader();
-      while ((line = reader.readLine()) != null) jb.append(line);
+      while ((line = reader.readLine()) != null) stringBuilder.append(line);
```
Tetheta, line 72: Also FYI I think Google's style guide prefers to always have {} when using while/if/etc statements. Just for added clarity.

PR #22, opened 2020-08-06
Restyles Form

hailed748:
image


Tetheta:
Starting review

src/main/webapp/courseLoader.js
```diff
@@ -1,127 +1,200 @@
+let termData;
```
Tetheta, line 1: This is a global which is a bit scary. Are you sure you need this to be global?

If anyone redeclares termData anywhere in the page's other JavaScript (or the user redefines window.termData) this will get changed.

Tetheta: If you need it for this whole file, consider using an IIFE https://flaviocopes.com/javascript-iife/ using the arrow function variety means the rest of the code shouldn't need to be changed, just wrapped inside of it.

src/main/webapp/courseLoader.js
```diff
@@ -1,127 +1,200 @@
...
-  fetch(`/term-live?school-name=${schoolName}&course-name=${courseName}&term=${term}&prof-name=${profName}&num-units=${units}`)
+  fetch(
+    `/term-live?school-name=${schoolName}&course-name=${courseName}&term=${term}&prof-name=${profName}&num-units=${units}`
+  )
     .then((response) => response.json())
-    .then((data => {
-        makeGraphs(data);
-    }));
+    .then((data) => {
+      termData = data;
```
Tetheta, line 27: This is the only usage of termData I see, why not just pass it through directly? Was this a test only variable?

src/main/webapp/courseLoader.js
```diff
@@ -1,127 +1,200 @@
...
-          hideBucketItems: true
-        }
-      };
-    const profPerceptionChart = new google.visualization.Histogram(document.getElementById("prof-chart"));
-    profPerceptionChart.draw(profPerceptionData, profPerceptionOptions);
+document.addEventListener("DOMContentLoaded", function () {
+  populateData();
+});
+
+function createComment(commentText, flag) {
```
Tetheta, line 36: flag isn't terribly helpful. what kind of flag is this? What is it determining? Based on the code maybe isCourse would be a good name?

src/main/webapp/courseLoader.js
```diff
@@ -1,127 +1,200 @@
...
+    createComment(comment, flag);
+  }
+}
+
+function makeGraphs(dataObject) {
+  const termCommentsList = dataObject.termCommentsList;
+  const dummyComments = ["dummy comment 1", "dummy comment 2"].concat(
+    termCommentsList
+  );
+  loadComments(dummyComments, /* true means course comments */ true);
```
Tetheta, line 71: The comments here on true/false are good. Even better would be a more descriptive flag "isCourse" and just putting that name next to the boolean here.

src/main/webapp/courseLoader.js
```diff
@@ -1,127 +1,200 @@
...
+  commentContainer.appendChild(commentWrapper);
 }
 
+function loadComments(commentList, flag) {
+  for (let comment of commentList) {
+    createComment(comment, flag);
+  }
+}
+
+function makeGraphs(dataObject) {
```
Tetheta, line 66: Not thrilled with the "dataObject" name. It does tell you the structure (object) and that it contains data, but absolutely nothing about that data.
termData or termDataObject would be better imo.

src/main/webapp/courseLoader.js
```diff
@@ -1,127 +1,200 @@
...
-    url.searchParams.set("hourOfWork", hours);
-    url.searchParams.set("difficulty", diff);
-    url.searchParams.set("term-input", termInput);
-    url.searchParams.set("prof-input", profInput);
-    url.searchParams.set("rating-term", ratingTerm);
-    url.searchParams.set("rating-prof", ratingProf);
-    fetch(url, { method: "POST" });
-}
\ No newline at end of file
+function passData() {
```
Tetheta, line 166: FYI this is going to conflict with Laura's PR, since she's doing the TODO.

Tetheta:
Design/etc looks clean btw, nice work on that.

src/main/webapp/courseLoader.js
hailed748: Okay this is helpful because in a later PR I chose to just store variables like schoolName, termName, etc as global because they don't change and I uses them everywhere.

src/main/webapp/courseLoader.js
hailed748: I left it like this for the future because I envisioned a case where I would need it elsewhere, but for the time being I will remove it until I need it.

src/main/webapp/courseLoader.js
hailed748: Yah I think we are both aware of this, whose ever PR gets submitted latest will have to sort the merge issues. Or honestly I can remove it from this PR because the main focus of this PR was the graph updates. Is there anything wrong with that approach?

Tetheta: Nope, that sounds fine.

PR #23, opened 2020-08-10
Adds term comparison graph

hailed748:
Adds a couple of functions to create a graphic comparing the active terms average rating, to that of the two previous terms.

Note: I added dummy data for the purpose of visualization, that data will be removed once linked with Laura's new code.

image


src/main/webapp/courseLoader.js
```diff
@@ -1,164 +1,205 @@
...
+      histogram: {
+        hideBucketItems: true,
+      },
+    };
+    const termPerceptionChart = new google.visualization.Histogram(
+      document.getElementById("term-chart")
+    );
+    termPerceptionChart.draw(termPerceptionData, termPerceptionOptions);
+
+    makeTermRatingChart();
```
ninaprabhu, line 109: Nit: Do you have to await this? I honestly have no idea so if it works right now then it's probably fine, but figured I'd point out

hailed748: I probably should, but it is working right now

Tetheta:
Starting review

src/main/webapp/courseLoader.js
Tetheta: It works now because you don't need to have anything happen after the async function returns/etc.

However it's good practice to always return the async promise if you're doing something async, so I'd return this call instead of just calling it.

That way if whatever calls makeGraphs() in the future needs to do something afterwards it can just await or .then() it.

This also makes testing easier if you're testing the public API but the private function happens to be async.

src/main/webapp/courseLoader.js
```diff
@@ -1,164 +1,205 @@
-google.charts.load("current", { packages: ["corechart"] });
+(() => {
+  google.charts.load("current", { packages: ["corechart"] });
+  google.charts.load("current", { packages: ["bar"] });
 
-function fillTitles() {
+  let termData;
```
Tetheta, line 5: I still don't think you need this to be script-wide. The only places you use it you either already pass in the data or could easily pass it in. 

Up to you if you want to change it. I've just had a lot of nightmares with tracking state like this and I prefer stateless code whenever possible.

src/main/webapp/courseLoader.js
```diff
@@ -1,164 +1,205 @@
...
+    const termPerceptionChart = new google.visualization.Histogram(
+      document.getElementById("term-chart")
+    );
+    termPerceptionChart.draw(termPerceptionData, termPerceptionOptions);
+
+    makeTermRatingChart();
+  }
+
+  async function makeTermRatingChart() {
+    const currentTerm = term.split(/(\s+)/).filter((e) => e.trim().length > 0);
```
Tetheta, line 113: Good to add a comment to any regex used, even if it's pretty simple like this one.

Tetheta, line 113: I see that you check the trimmed length but don't actually update the values to be trimmed. Might want to first map and trim, then filter if you want things trimmed.

src/main/webapp/courseLoader.js
```diff
@@ -1,164 +1,205 @@
...
+    let prevTerm1;
+    let prevTerm2;
+
+    if (currentTerm[0] == "Spring") {
+      prevTerm1 = `Fall ${currentTerm[1]}`;
+      prevTerm2 = `Spring ${String(parseInt(currentTerm[1] - 1))}`;
+    } else {
+      prevTerm1 = `Spring ${currentTerm[1]}`;
+      prevTerm2 = `Fall ${String(parseInt(currentTerm[1] - 1))}`;
+    }
```
Tetheta, line 123: Note that this doesn't handle summer classes, or schools on the quarter system (where there're spring, summer, fall, and winter terms)

Tetheta: (not that this is an issue for MVP, just making sure you're aware)

src/main/webapp/courseLoader.js
```diff
@@ -1,164 +1,205 @@
...
+
+    const prevTermData1 = await getPreviousTermData(prevTerm1);
+    const prevTermData2 = await getPreviousTermData(prevTerm2);
```
Tetheta, line 126: Right now this is going to fetch one of them at a time.

I'd instead await on Promise.all so you can fetch both at once: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

src/main/webapp/courseLoader.js
```diff
@@ -1,164 +1,205 @@
...
+
+    const average = (list) =>
+      list.reduce((prev, curr) => prev + curr) / list.length;
+
```
Tetheta, line 130: Add an initial value to this reduce statement (second arg with just 0 or w/e).
Otherwise it'll throw an error when you operate on an empty array

src/main/webapp/courseLoader.js
hailed748: Yah I have been thinking of a good way to factor in all systems without having a ton of if/else, especially due to the variety in terms per school

hailed748: I added a TODO until I figure out a good way

src/main/webapp/courseLoader.js
```diff
@@ -122,14 +125,21 @@
       prevTerm2 = `Fall ${String(parseInt(currentTerm[1] - 1))}`;
     }
 
-    const prevTermData1 = await getPreviousTermData(prevTerm1);
-    const prevTermData2 = await getPreviousTermData(prevTerm2);
+    let prevTermData1;
+    let prevTermData2;
+    await Promise.all([
```
Tetheta, line 130: Is this actually doing what you want? Interesting if so.

Typically await and .then() are used separately.

For example, here I'd probably do something like:
```
const [prevTermData1, prevTermData2] = await Promise.all([getPreviousTermData(prevTerm1), getPreviousTermData(prevTerm2)]);
```
(This is using the array destructuring shorthand to make things cleaner)

This current implementation might end up working because the .then waits the same amount of time here as the await, so the variables will be set. But typically the .then() would happen after the rest of the function executed. So it's weird to have it in this format.

hailed748: Yah I think I need to read up on the await, promise, and async stuff. I feel like there is a decent gap in my knowledge on this topic.

PR #25, opened 2020-08-11
Adding user ID to DataServlet

lqueipom:
Modified DataServlet to add a user ID to the Rating Entity. 

Added verify() and onLoad() to pass the user ID to the DataServlet.


src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -1,22 +1,13 @@
-// Copyright 2019 Google LLC
```
ninaprabhu, line 1: Could you potentially git rm this file if you push to master before me? Or pull from master before pushing if I push before you? Just wanna make sure we both have the same version of this file to avoid this happening again haha

src/main/webapp/login-script.js
```diff
@@ -1,25 +1,35 @@
-function onSignIn(googleUser) {
+async function onSignIn(googleUser) {
```
ninaprabhu, line 1: this file is also probably something to git rm or otherwise make sure is the same between us

src/main/webapp/search-script.js
```diff
@@ -1,8 +1,8 @@
 function collectUnitsFromSearch() {
```
ninaprabhu, line 1: ok maybe this is an issue that will go away when I push? Maybe this is something prettier is doing

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -63,28 +61,51 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    String professorFeedback = request.getParameter("prof-input");
-    Long professorRating = Long.parseLong(request.getParameter("rating-professor"));
+
+    StringBuilder jb = new StringBuilder();
+    String line = null;
+    try {
+      BufferedReader reader = request.getReader();
+      while ((line = reader.readLine()) != null) jb.append(line);
+    } catch (Exception exception) {
+      /*report an error*/
```
hailed748, line 75: would you now want to throw the caught exception?

lqueipom:
> Kinda a confusing PR I think some of Nina's stuff was in here as well

Yeah! Worked with Nina on this so I had to make some changes on the file I had of hers. But either way will delete those from this commit to avoid possible merge conflicts later :)

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: Yes! Sorry, fixed this on my other PR will implement those changes as well here.

Tetheta:
Starting review

src/main/webapp/courseLoader.js
```diff
@@ -27,6 +27,198 @@ function populateData() {
...
+async function verify() {
+  const auth2 = gapi.auth2.getAuthInstance();
+  const googleUser = auth2.currentUser.get();
+  const token = googleUser.getAuthResponse().id_token;
+  const url = new URL("/login", window.location.origin);
+  url.searchParams.set("token", token);
+  const response = await fetch(url, { method: "POST" });
+  const userInfo = await response.json();
+  const id = userInfo.id;
+  return id;
```
Tetheta, line 149: might as well just return userInfo.id here instead of making a new id variable

PR #28, opened 2020-08-12
Finds Previous Terms & Adds Perception Chart

hailed748:


src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -162,4 +167,31 @@ private void putNewEntities(Entity school, Entity course, Entity term, Entity pr
...
+      startDay.set(Calendar.DATE, 01);
+      startDay.set(Calendar.YEAR, termYear);
+    } else if (termList[0].equals("Fall")) {
+      startDay.set(Calendar.MONTH, 8);
+      startDay.set(Calendar.DATE, 01);
+      startDay.set(Calendar.YEAR, termYear);
+    } else {
+      startDay.set(Calendar.MONTH, 11);
+      startDay.set(Calendar.DATE, 01);
+      startDay.set(Calendar.YEAR, termYear);
```
ninaprabhu, line 192: Is it better style to create a map or something and make a function for this?

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,123 @@
...
+    Query termQuery =
+        new Query("Term")
+            .setAncestor(courseKey)
+            .addSort("timeStamp", SortDirection.DESCENDING)
+            .setFilter(timeFilter);
+    List<Entity> foundTerms = db.prepare(termQuery).asList(FetchOptions.Builder.withLimit(2));
+    return foundTerms;
+  }
+
+  private Entity getTerm(DatastoreService db, HttpServletRequest request) {
```
lqueipom, line 68: Do you need this function? Could you not directly call findTerm(db, request.getParameter("school-name")...)? If not, maybe change the name of the function as by name getTerm and findTerm looks like they do the same.

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,123 @@
...
+    return foundTerm;
+  }
+
+  private Entity findTerm(
+      DatastoreService db,
+      String schoolName,
+      String courseName,
+      String termName,
+      Long units,
+      String profName) {
```
lqueipom, line 85: Do you use profName at all in this function?

src/main/webapp/course.html
```diff
@@ -134,7 +129,7 @@ <h1 class="p-4 text-center">Add Your Review</h1>
               <p class="lead">Hours/Week on this Course</p>
               <input
                 id="hours-work"
-                name="hoursOfWork"
+                name="hours-work"
```
lqueipom, line 132: changed this to hours in my version, so for consistency do you want me to change to hours-work or viceversa?

lqueipom: I mean id="hours"

hailed748: I can change it to hours

src/main/java/com/google/sps/servlets/AddSchoolData.java
hailed748: Not sure what you mean by map, we should discuss it during our standup.

hailed748: The hashmap was causing me some trouble so I just simplified the code as best as I could for the time being. This was laura can test her stuff

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<Entity> prevTermsData = getPreviousTerms(db, request);
+    String prevTermsDataJSON = makeJSON(prevTermsData);
+    response.setContentType("application/json;");
+    response.getWriter().println(prevTermsDataJSON);
+  }
+
+  public List<Entity> getPreviousTerms(DatastoreService db, HttpServletRequest request) {
```
ninaprabhu, line 36: does this need to be public?

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private List<Entity> findQueryMatch(
+      DatastoreService db, String entityType, String entityProperty, String propertyValue) {
+    Filter filter = new FilterPredicate(entityProperty, FilterOperator.EQUAL, propertyValue);
+    Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
+    return result;
+  }
+
+  private String makeJSON(Object changeItem) {
```
ninaprabhu, line 102: do you ever check that this fails (and if so, do you do something)? I'm just seeing that you use the return value indiscriminately

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,112 @@
...
+    Key courseKey =
+        db.prepare(courseQuery).asList(FetchOptions.Builder.withDefaults()).get(0).getKey();
+
+    Filter timeFilter = new FilterPredicate("timeStamp", FilterOperator.LESS_THAN, startTime);
+    Query termQuery =
+        new Query("Term")
+            .setAncestor(courseKey)
+            .addSort("timeStamp", SortDirection.DESCENDING)
+            .setFilter(timeFilter);
+    List<Entity> foundTerms = db.prepare(termQuery).asList(FetchOptions.Builder.withLimit(2));
```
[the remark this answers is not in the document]
LukeGMTravis: Looking at the client side it looks like this is the case. An improvement here might also be to let the client request a maximum number of past terms. That would make your API as flexible as possible and prevent hardcoding numbers like this.

src/main/webapp/courseLoader.js
```diff
@@ -167,11 +133,66 @@
     chart.draw(comparisonData, google.charts.Bar.convertOptions(options));
   }
 
+  async function makeTermPerceptionChart(termDataObject) {
+    const [prevTerm1, prevTerm2] = await getPrevTermName();
```
LukeGMTravis, line 137: Ideally you'd probably want this to be flexible and adapt to any number of returned results, not just arbitrarily 2. This makes things much easier if/when you decide you want 3. Or 4. Or 5. Ect. Web dev is full of iteration so it's best to code with that in mind : )

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -162,4 +165,26 @@ private void putNewEntities(Entity school, Entity course, Entity term, Entity pr
...
+    startDay.set(Calendar.YEAR, termYear);
+    startDay.set(Calendar.DATE, 01);
+
+    if (termList[0].equals("Spring")) {
+      startDay.set(Calendar.MONTH, 2);
+    } else if (termList[0].equals("Summer")) {
+      startDay.set(Calendar.MONTH, 5);
+    } else if (termList[0].equals("Fall")) {
+      startDay.set(Calendar.MONTH, 8);
+    } else {
```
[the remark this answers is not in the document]
hailed748: Good point. Also might turn the form into a drop down so there can be a good chance of no errors.

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,112 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<Entity> prevTermsData = getPreviousTerms(db, request);
+    String prevTermsDataJSON = makeJSON(prevTermsData);
+    response.setContentType("application/json;");
+    response.getWriter().println(prevTermsDataJSON);
+  }
+
+  private List<Entity> getPreviousTerms(DatastoreService db, HttpServletRequest request) {
+    String schoolName = request.getParameter("school-name");
+    String courseName = request.getParameter("course-name");
```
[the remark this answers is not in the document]
hailed748: Yes so I discussed this with Nina today and we think it is feasible to make this change. Laura was not able to make our standup but we will discuss with her to see what impact this will have on her code as well, but based on the discuss Nina and I had we think it can be done with minor adjustments.

src/main/webapp/courseLoader.js
hailed748: modified this so now it works for all numbers of comparisons. Currently I leave it hard coded for 2 previous terms, but I will eventually add a feature on to let them select how many they want compare across.

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,112 @@
...
+    Query courseQuery =
+        new Query("Course").setAncestor(schoolKey).setFilter(CompositeFilterOperator.and(filters));
+    Key courseKey =
+        db.prepare(courseQuery).asList(FetchOptions.Builder.withDefaults()).get(0).getKey();
+
+    Filter termFilter = new FilterPredicate("term", FilterOperator.EQUAL, termName);
+    Query termQuery = new Query("Term").setAncestor(courseKey).setFilter(termFilter);
+    Entity foundTerm = db.prepare(termQuery).asList(FetchOptions.Builder.withDefaults()).get(0);
+
+    return foundTerm;
```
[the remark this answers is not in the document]
hailed748: One other quick question. In my PR you commented  my functions needing to return NOT_FOUND if a query comes back empty, but we are guaranteed that this can't be the case because these functions are called from a course page and we can only reach that course page if their corresponding entities exists in the DB. If this is the case should I still have a NOT_FOUND case or just leave a comment?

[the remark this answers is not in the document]
hailed748: Added a comment for the time being. When we shift to key/id in the url I will change it to throw a Datastore Note found error.

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -162,4 +165,28 @@ private void putNewEntities(Entity school, Entity course, Entity term, Entity pr
...
+  private Date findTermDate(String termName) {
+    String[] termList = termName.split(" ");
+
+    int month = 0;
+    int termYear = Integer.parseInt(termList[1]);
+    if (termList[0].equals("Spring")) {
+      month = 2;
+    } else if (termList[0].equals("Summer")) {
+      month = 5;
+    } else if (termList[0].equals("Fall") | termList[0].equals("Autumn")) {
```
LukeGMTravis, line 178: The Java or operator is `||`

src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -162,4 +165,28 @@ private void putNewEntities(Entity school, Entity course, Entity term, Entity pr
...
+      month = 8;
+    } else if (termList[0].equals("Winter")) {
+      month = 0;
+    } else {
+      throw new IllegalArgumentException();
+    }
+
+    Calendar startDay = Calendar.getInstance();
+    startDay.set(termYear, month, 01);
```
LukeGMTravis, line 187: You don't need the `0` in `01`

src/main/java/com/google/sps/servlets/PreviousTerms.java
```diff
@@ -0,0 +1,114 @@
...
+    String prevTermsDataJSON = makeJSON(prevTermsData);
+    response.setContentType("application/json;");
+    response.getWriter().println(prevTermsDataJSON);
+  }
+
+  private List<Entity> getPreviousTerms(DatastoreService db, HttpServletRequest request) {
+    String schoolName = request.getParameter("school-name");
+    String courseName = request.getParameter("course-name");
+    Long units = Long.parseLong(request.getParameter("num-units"));
+    Integer comparisonCount = Integer.parseInt(request.getParameter("comparison-count"));
```
LukeGMTravis, line 41: "comparison-count" implies that you're making comparisons but really it seems more accurate to say it's a limit for how many previous terms you return. I'd recommend renaming to "term-limit" or something like that

src/main/java/com/google/sps/servlets/PreviousTerms.java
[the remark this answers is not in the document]
LukeGMTravis: Sounds good

src/main/webapp/courseLoader.js
```diff
@@ -81,97 +92,103 @@
...
     const response = await fetch(url);
+    return response.json();
+  }
+
+  async function getPrevTermName(comparisonCount) {
+    const url = `/prev-terms?school-name=${schoolName}&course-name=${courseName}&term=${term}&prof-name=${profName}&num-units=${units}&comparison-count=${comparisonCount}`;
+    const response = await fetch(url);
     const prevTermData = await response.json();
-    return prevTermData;
+    const [prevTermName1, prevTermName2] = [
```
LukeGMTravis, line 187: This is still hard coding that you'll only ever deal with 2 returned terms. You can make this more flexible by doing something like:
```
return prevTermData.map((data) => data.properties.term);
```

It's shorter this way too : )

hailed748: oh this is a bug forgot to change this to just a list

hailed748: Nice catch I forgot to change the return!

PR #29, opened 2020-08-12
Fuzzy search

ninaprabhu:


src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -41,18 +42,26 @@
...
-    String coursesJson = new Gson().toJson(courses);
+    // String resultsJson = new Gson().toJson(results);
     response.setContentType("application/json;");
-    response.getWriter().println(coursesJson);
+    response.getWriter().println(results);
   }
 
   /* Create list of courses given request. Used for testing. */
-  public List<Course> getHelper(HttpServletRequest request) {
+  public JSONObject getHelper(HttpServletRequest request) {
```
lqueipom, line 53: Maybe change getHelper to a name that better describes the function :)

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -41,18 +42,26 @@
...
+    String success = "good";
     List<Filter> filters = getFilters(request);
-    List<Entity> results = getResults(filters);
-    List<Course> courses = getCourses(results);
+    List<Entity> results = getResults(filters, false);
+    if (results.size() == 0) {
+      results = getResults(filters, true);
```
hailed748, line 58: Would you not have to check that the results is not empty before you can say success or am I misunderstanding the code?
`if(results.isEmpty()){`
  `  results = getResults(filters,true);`
  `  if(!results.isEmpty()){`
     `    success = "okay"`
 `   }`
`}`

hailed748, line 58: use isEmpty()

ninaprabhu: Success represents how successful the search was (exact match  = good, similar matches = okay, no matches = bad). I use it to print a message when displaying results, if applicable.

ninaprabhu:
Screenshot 2020-08-13 at 12 59 50 PM

Tetheta:
Starting review

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -41,27 +42,44 @@
...
   /* Create list of filters given parameters specified in request. */
-  private List<Filter> getFilters(HttpServletRequest request) {
+  private List<Filter> getFilters(HttpServletRequest request, boolean fuzzy) {
     List<Filter> filters = new ArrayList<>();
     if (!request.getParameter("courseName").isEmpty()) {
       String name = request.getParameter("courseName");
-      Filter nameFilter = new FilterPredicate("name", FilterOperator.EQUAL, name);
+      Filter nameFilter;
+      if (fuzzy) {
+        String dept = name.split(" ")[0];
```
Tetheta, line 74: Another point where shortening the name doesn't really make sense. I'm guessing this is department, but it could also be depot, or a few other things.
Prefer more descriptive names in general, that way I don't have to remember context to remember what a variable is referring to.

Tetheta: Also it removes any second guessing when debugging something. Having in the back of your mind "I assumed it meant department, but what if it didn't?" takes up brain space and can delay a solution.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -73,11 +91,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+        terms.add(getNextTerm(term));
+        termFilter = new FilterPredicate("term", FilterOperator.IN, terms);
+      } else {
+        termFilter = new FilterPredicate("term", FilterOperator.EQUAL, term);
+      }
       filters.add(termFilter);
     }
 
-    if (!request.getParameter("units").isEmpty()) {
+    if (!request.getParameter("units").isEmpty() && !(fuzzy)) {
```
Tetheta, line 106: Shouldn't need () around fuzzy here I don't think unless Java is being extra weird

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -147,4 +185,25 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  }
+
+  private String getNextTerm(String term) {
+    String season = term.split(" ")[0];
+    int year = Integer.parseInt(term.split(" ")[1]);
+    if (season.equals("Fall")) {
+      return "Spring " + String.valueOf(year + 1);
+    } else { // Season is "Spring"
+      return "Fall " + String.valueOf(year);
+    }
```
Tetheta, line 207: You and Dagm should get together on this code and separate it out into a utility.
Whatever you end up doing, you'll want to be consistent.

src/main/webapp/search-script.js
```diff
@@ -107,7 +111,7 @@ async function verify() {
     const url = new URL("/login", window.location.origin);
     url.searchParams.set("token", token);
     const response = await fetch(url, {method:"POST"});
-    const userInfo = await response.json();
+    const userInfo = response.json();
```
Tetheta, line 114: Why remove the await? Is response.json() actually not async here?

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -41,27 +42,44 @@
   @Override
   /* Show courses. */
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Course> courses = getHelper(request);
-
-    String coursesJson = new Gson().toJson(courses);
+    JSONObject results = getMatchingCourses(request);
```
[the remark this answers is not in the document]
ninaprabhu: I changed it to results because it holds courses and a message, but I think you're right in that courseResults is still a better name than results in this case. Thanks!

src/main/webapp/search-script.js
ninaprabhu: I thought it wasn't, but I checked online and it definitely is haha, changed back

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: In Dagm's JS he's currently roughly doing the same thing as me (I think he's just getting the date of a term to use as a timestamp), but will sync up with him on this to make sure we don't have any overlapping code

src/main/java/com/google/sps/data/Course.java
```diff
@@ -1,19 +1,38 @@
...
+      String termName,
+      String schoolName,
+      DatastoreService db) {
     name = courseName;
     professor = professorName;
     units = numUnits;
     term = termName;
     school = schoolName;
+    termKey = findTermKey(db, schoolName, courseName, termName, units, professorName);
+    // termKey = null;
```
hailed748, line 35: remove comment

src/main/webapp/search-script.js
```diff
@@ -46,14 +50,11 @@ function createListElement(course) {
   const liElement = document.createElement('li');
   const link = document.createElement('a');
   const url = new URL("/course.html", window.location.origin);
-  url.searchParams.set("course-name", course.name);
-  url.searchParams.set("prof-name", course.professor);
-  url.searchParams.set("num-units", course.units);
-  url.searchParams.set("term", course.term);
+  url.searchParams.set("term0key", course.termKey);
```
hailed748, line 53: I think this is a typo and you meant "term-key"

Tetheta:
Starting review

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
+function collectUnitsFromSearch() {
+  const inputElems = document.getElementsByName("search-units");
+  const addedUnits = []
```
Tetheta, line 3: Nit: add a semi-colon

Seems that prettier didn't get run somehow? Should've done that for you

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+  for (let elem of inputElems) {       
+    if (elem.checked){
```
Tetheta, line 5: Space before {, seems prettier isn't running

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+
+function collectUnitsFromInput() {
+  const inputElems = document.getElementsByName("num-units");
+  const addedUnits = []
+  for (let elem of inputElems) {       
+    if (elem.checked){
+      addedUnits.push(elem.value);
+    }
+  }
+  return addedUnits;
```
Tetheta, line 20: Seems everything but the inputElems is identical for these two functions. Consider creating a utility function instead that takes in the inputElems as a parameter

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+  url.searchParams.set("courseName", courseName);
+  url.searchParams.set("profName", profName);
+  url.searchParams.set("units", units);
+  url.searchParams.set("term", termName);
+  url.searchParams.set("schoolName", school)
+  fetch(url)
+  .then(response => response.json())
+  .then((searchResults) => {
+    if (searchResults.hasOwnProperty("message")) { // TODO: Make a div for this message so specific styling is easier.
+        courseResults.innerHTML += searchResults.message + "<br />";
```
Tetheta, line 41: FYI this allows for XSS injection pretty easily, as hackers can change the searchResults.message variable to their own script tag and do whatever they want with host page permissions.

Don't worry about it for now, but if we were making a real website this would be forbidden (and we'd have protected methods for accessing innerHTML/etc).

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+  return liElement;
+}
+
+function getUserSchool() {
+  const auth2 = gapi.auth2.getAuthInstance();
+  const profile = auth2.currentUser.get().getBasicProfile();
+  const email = profile.getEmail();
+  const start = email.indexOf('@');
+  const end = email.lastIndexOf('.');
+  const school = email.substring(start+1, end);
```
Tetheta, line 69: can just return this email.substring, no need for school variable

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+  const termName = document.getElementById("term").value;
+  const units = collectUnitsFromInput();
+  const schoolName = getUserSchool();
+  const url = new URL("/search", window.location.origin);
+  url.searchParams.set("course-name", courseName);
+  url.searchParams.set("prof-name", profName);
+  url.searchParams.set("num-units", units);
+  url.searchParams.set("term", termName);
+  url.searchParams.set("school-name", schoolName);
+  fetch(url, {method:"POST"});
```
Tetheta, line 85: Good practice to return this in case we need to chain it later

Tetheta: Also if you're using async await elsewhere, I'd suggest changing this to it too and await on the fetch.

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+  const url = new URL("/search", window.location.origin);
+  url.searchParams.set("courseName", courseName);
+  url.searchParams.set("profName", profName);
+  url.searchParams.set("units", units);
+  url.searchParams.set("term", termName);
+  url.searchParams.set("schoolName", school)
+  fetch(url)
+  .then(response => response.json())
+  .then((searchResults) => {
+    if (searchResults.hasOwnProperty("message")) { // TODO: Make a div for this message so specific styling is easier.
```
Tetheta, line 40: Since you're using async await later, might be good to convert this to it as well. Typically good to keep the file as consistent as possible with using async/await or avoiding it.

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+async function onSignIn(googleUser) {
+  const profile = googleUser.getBasicProfile();
+  const token = googleUser.getAuthResponse().id_token;
+  const url = new URL("/login", window.location.origin);
+  url.searchParams.set("token", token);
+  const response = await fetch(url, {method:"POST"});
+  const id = await response.json();
+    if (id.verified) { // Successful sign-in.
```
Tetheta, line 95: Don't need indents.

If you `npm install -g prettier` in terminal, you should be able to run `prettier /path/to/this/file.js` to format it manually if the automatic thing isn't working.

src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -152,4 +284,12 @@ private void assertEqualsCourseArrays(List<Course> courses, List<Course> expecte
       assertEquals(actual.getSchool(), expected.getSchool());
     }
   }
+
+  private List<Entity> findQueryMatch(
+      DatastoreService db, String entityType, String entityProperty, String propertyValue) {
+    Filter filter = new FilterPredicate(entityProperty, FilterOperator.EQUAL, propertyValue);
+    Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
```
Tetheta, line 292: Just return this, no need for a result variable

Tetheta:
Might be some of the comments are on stuff from other PRs, got a lot in this PR. 
Looking good overall, lot of little nits and tweaks

src/main/webapp/search-script.js
ninaprabhu: It seems like prettier just isn't formatting my .js file, I'll go through and manually format it

src/main/webapp/search-script.js
ninaprabhu: ooh that's so interesting, had no idea. I'll look into that!

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -103,48 +139,121 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    json.put("courses", strCourses);
+    return json;
   }
 
   /* Store course. */
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+    addCourse(request, db);
+    response.sendRedirect("/index.html");
```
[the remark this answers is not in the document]
ninaprabhu: Oh for some reason I thought it went to the servlet route page unless I sent a redirect. Thanks!

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -41,27 +44,51 @@
   @Override
   /* Show courses. */
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Course> courses = getHelper(request);
-
-    String coursesJson = new Gson().toJson(courses);
```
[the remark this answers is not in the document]
ninaprabhu: I do it inside of the helper function now

src/main/java/com/google/sps/data/Course.java
```diff
@@ -35,4 +56,54 @@ public String getTerm() {
...
+    Entity foundTerm = db.prepare(termQuery).asList(FetchOptions.Builder.withDefaults()).get(0);
+    List<Key> keys = Arrays.asList(foundTerm.getKey(), courseKey);
+    return keys;
+  }
+
+  private List<Entity> findQueryMatch(
+      DatastoreService db, String entityType, String entityProperty, String propertyValue) {
+    Filter filter = new FilterPredicate(entityProperty, FilterOperator.EQUAL, propertyValue);
+    Query q = new Query(entityType).setFilter(filter);
+    List<Entity> result = db.prepare(q).asList(FetchOptions.Builder.withDefaults());
```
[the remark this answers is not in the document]
ninaprabhu: Seems like this is something I do a lot - I thought it was better style bc the variable name is clear but definitely see how it's useless. Sorry for any annoyance/confusion haha

src/main/java/com/google/sps/data/Course.java
```diff
@@ -1,19 +1,40 @@
...
+      String schoolName,
+      DatastoreService db) {
     name = courseName;
     professor = professorName;
     units = numUnits;
     term = termName;
     school = schoolName;
+    List<Key> keys = findKeys(db, schoolName, courseName, termName, units, professorName);
+    termKey = keys.get(0);
+    courseKey = keys.get(1);
```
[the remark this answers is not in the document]
ninaprabhu: Dagm and I use the findTermKey and findCourseKey functions in the same way  - we only call them on courses that have been successfully added to Datastore, so we know them (and their keys) exist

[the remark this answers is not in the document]
ninaprabhu: Should I put some kind of default message or a try/catch statement even if we're certain they'll never error out though (because I'm sure there's probably some kind of scenario in which we'd be wrong haha)

src/main/java/com/google/sps/data/Course.java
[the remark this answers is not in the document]
Tetheta: That's not invalid. Particularly in JavaScript that can often be helpful since there's no typing on the return statement to clue you into what's being passed back. 

It just is usually redundant in Java, and often the function name should tell you what it's returning anyways. 
It also is a very slight memory increase (extra variable stored) but so slight it would never be noticeable on benchmarks lol. Just more the principle of the thing.

src/main/java/com/google/sps/data/Course.java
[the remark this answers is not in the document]
Tetheta: Eh I'd just let it fail then. If you're pretty sure it's never wrong, it's better to make it fail catastrophically, that way you know for sure if your assumptions were invalid.

src/main/webapp/search-script.js
Tetheta: Yeah it's a super unintuitive API ._.

PR #34, opened 2020-08-17
Bug fix

hailed748:
Fixes a bug in which separate course entities were not being created for courses with the same name but different units.


Tetheta:
Starting review

PR #36, opened 2020-08-17
Populates Course Title using TermKey Parameter

hailed748:
Refactors code so now the course name, term, and number of students enrolled is populated via the term key.

Adds: term info servlet that returns the information mentioned above from the new term-key parameter being passed in the url


src/main/java/com/google/sps/servlets/AddSchoolData.java
```diff
@@ -7,10 +7,12 @@
 import com.google.appengine.api.datastore.FetchOptions;
```
ninaprabhu, line 7: consider merging this into your class count PR branch because I think the changes to AddSchoolData.java and AddSchoolData.html are in both PRs

Tetheta:
Starting review

PR #37, opened 2020-08-17
Refactors LiveCourseData & CourseLoader

hailed748:
Refactors LiveCourseData servlet to work with the new params of term-key and course-key. The move to keys as params has greatly reduced the code needed in these files.

Adds: PreviousTermKey servlet that gets the key corresponding to each previous term, this is then feed into the LiveCourseData servlet which gets all the data

### DIFF
--- a/src/main/java/com/google/sps/servlets/PreviousTermKey.java
+++ b/src/main/java/com/google/sps/servlets/PreviousTermKey.java
@@ -6,6 +6,7 @@ import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Query;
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/sps/servlets/PreviousTermKey.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/prev-key")
+public class PreviousTermKey extends HttpServlet {
+  private DatastoreService db = DatastoreServiceFactory.getDatastoreService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
ninaprabhu, line 24: is the point of this servlet to get all previous terms a course was taught in? or just one previous term

ninaprabhu: Could you just add a few comments to this servlet in general? I'm a little confused about what's going on/if the term object servlet I'm making rn overlaps with this

PR #39, opened 2020-08-18
Implementing Translation

lqueipom:
Added: 
- Funcion that translates feedback in DataServlet.
- Added field to form for user to decide whether they would like to translate feedback.

Changed: 
- Changed in courseLoader.js what parameters I get from the URL
- Changed in DataServlet how I find my term key (now it is passed directly from URL)


src/main/webapp/courseLoader.js
```diff
@@ -238,27 +237,31 @@
   async function getRatingPropertiesToStore() {
     const queryString = window.location.search;
     const urlParams = new URLSearchParams(queryString);
+    let wantsToTranslate;
```
ninaprabhu, line 240: change this to just translate?

src/main/webapp/courseLoader.js
```diff
@@ -238,27 +237,31 @@
...
+
+    if (document.getElementById("true").checked) {
+      wantsToTranslate = document.getElementById("true").value;
+    } else if (document.getElementById("false".checked)) {
```
ninaprabhu, line 244: document.getElementById("false").checked

src/main/webapp/courseLoader.js
```diff
@@ -238,27 +237,31 @@
   async function getRatingPropertiesToStore() {
     const queryString = window.location.search;
     const urlParams = new URLSearchParams(queryString);
+    let wantsToTranslate;
+
+    if (document.getElementById("true").checked) {
```
ninaprabhu, line 242: why don't you just have one box that the user can check if they want to translate? that way you don't have to deal with a case where someone checks both or no boxes

src/main/webapp/courseLoader.js
```diff
@@ -271,11 +274,8 @@
   }
 
   function newURL(
```
ninaprabhu, line 276: it doesn't seem like you use the newURL function anywhere other than right above this, so why don't you only set one parameter for the function (whatever ratingProperties is) and then call each specific attribute inside the function itself

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,129 @@
+function collectUnitsFromSearch() {
```
ninaprabhu, line 1: wait why is this here? sorry if i'm missing something but does this need to be in your PR?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -123,7 +129,7 @@ public void addTermRating(HttpServletRequest request) throws IOException {
 
     Entity termRatingEntity =
         termRatingQueryList.isEmpty()
-            ? new Entity("Rating", currentTermKey)
+            ? new Entity("Rating", KeyFactory.stringToKey())
```
hailed748, line 132: you need to put the string key as a param in KeyFactory.stringToKey()

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -139,25 +145,23 @@ public void addTermRating(HttpServletRequest request) throws IOException {
...
         Document.newBuilder().setContent(feedback).setType(Document.Type.PLAIN_TEXT).build();
-    Sentiment sentiment = languageService.analyzeSentiment(feedbackDoc).getDocumentSentiment();
+    Sentiment sentiment = this.languageService.analyzeSentiment(feedbackDoc).getDocumentSentiment();
     float score = sentiment.getScore();
-    languageService.close();
+    // Won't be closing languageService as we want to use constructor.
     return score;
   }
 
   private Entity findTerm(
```
hailed748, line 156: I believe this function is not being used anymore so you can delete it

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -58,62 +69,57 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get written feedback.
-    addTermRating(request);
+    addTermRating(request, /* DatastoreService */ datastore);
```
hailed748, line 72: nit: probably don't need the comment because datastore is a good variable name

src/main/webapp/search-script.js
lqueipom: By accident deleted that file so had to add it back, sorry about that!

src/main/webapp/courseLoader.js
lqueipom: Decided on a dropdown menu instead so now I just have to call document.getElementById("translate").

LukeGMTravis:
Starting review

src/main/webapp/courseLoader.js
```diff
@@ -238,64 +237,41 @@
...
-      schoolName: urlParams.get("school-name"),
+      courseKey: urlParams.get("course-key"),
+      termKey: urlParams.get("term-key"),
       termInput: document.getElementById("term-input").value,
       profInput: document.getElementById("prof-input").value,
       ratingTerm: document.getElementById("rating-term").value,
       ratingProf: document.getElementById("rating-prof").value,
       hours: document.getElementById("hours").value,
       difficulty: document.getElementById("difficulty").value,
+      ID: await verify(),
```
LukeGMTravis, line 250: Nit: Acronyms are still typed in lowercase by convention (so this should just be "id")

src/main/webapp/search-script.js
```diff
@@ -0,0 +1,129 @@
...
+  const url = new URL("/search", window.location.origin);
+  url.searchParams.set("course-name", courseName);
+  url.searchParams.set("prof-name", profName);
+  url.searchParams.set("num-units", units);
+  url.searchParams.set("term", termName);
+  url.searchParams.set("school-name", schoolName);
+  fetch(url, { method: "POST" });
+}
+
+async function onSignIn(googleUser) {
```
LukeGMTravis, line 83: Nit: "onSignIn" sounds like the name of a callback parameter. This may be better named something along the lines of "signIn", "signInUser", "attemptSignIn", etc

LukeGMTravis: Since you have just "signOut" below then "signIn" likely makes the most sense.

PR #41, opened 2020-08-18
Refactors LiveCourseData & CourseLoader

hailed748:
Refactors LiveCourseData servlet to work with the new params of term-key and course-key. The move to keys as params has greatly reduced the code needed in these files.

Adds: PreviousTermKey servlet that gets the key corresponding to each previous term, this is then feed into the LiveCourseData servlet which gets all the data


src/main/java/com/google/sps/data/Course.java
```diff
@@ -1,19 +1,41 @@
 package com.google.sps.data;
```
ninaprabhu, line 1: just to clarify - did you change anything in this file?

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -18,6 +18,7 @@
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
```
ninaprabhu, line 18: didn't change this either right? not blocking or anything - just confirming so i don't have to compare my changes to this file

hailed748:
Having it go to master resolved a lot of the commits, but there are still some files in there that are from a previous PR.

hailed748:
Course.java, SearchServlet.java, search-script.js, SearchServletTest.java, TermInfo.java, AddSchoolData.java, AddSchoolData.html, and course.html are all files that have been reviewed before and are being added back due to deletion. Hope that makes reviewing things easier

Tetheta:
Starting review

hailed748:
In this case I actually wanted the search stuff because it got deleted from master so I am adding it back

PR #46, opened 2020-08-20
Integration

hailed748:
Resolves bugs that slipped through from change to term-key. Also removes a deprecated test.

### DIFF
--- a/src/main/webapp/search-script.js
+++ b/src/main/webapp/search-script.js
@@ -14,6 +14,7 @@ async function showCourses() {
   const courseName = document.getElementById("search-course").value;
   const profName = document.getElementById("search-prof").value;
   const termName = document.getElementById("search-term").value;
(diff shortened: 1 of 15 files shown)


LukeGMTravis:
Starting review

src/main/webapp/search-script.js
```diff
@@ -69,6 +69,7 @@ async function addCourse() {
   url.searchParams.set("num-units", units);
   url.searchParams.set("term", termName);
   url.searchParams.set("school-name", schoolName);
+  url.searchParams.set("num-enrolled", "300");
```
LukeGMTravis, line 72: 300 seems kinda arbitrary? Is this for mock data or testing or something?

hailed748: Yes that is temporary will be removed in future PR. We just need to be passing in something for the time being so it does not 500 multiple servlets.

PR #47, opened 2020-08-20
Adding Grades to Form

lqueipom:
Added:
- A grade selection dropdown in course.html.
- Required attribute to certain input fields in course.html.
- Bootstrap to style required fields and provide a custom message when fields are empty.
- Corresponding modifications in courseLoader.js and DataServlet so grades are now passed in as a Term Rating Entity Property.


src/main/webapp/course.html
```diff
@@ -163,18 +186,34 @@ <h1 class="p-4 text-center">Add Your Review</h1>
...
+              Please input a valid difficulty score.
+            </div>
+            <div class="col">
+              <p class="lead">What was your grade in the class?</p>
+              <select id="grade" name="grade" class="form-control mb-5">
+                <option value="A">A</option>
+                <option value="B">B</option>
+                <option value="C">C</option>
+                <option value="D">D</option>
+                <option value="E">E</option>
```
hailed748, line 204: Could you add A+, A-. Also I don't think there is an E letter grade.

LukeGMTravis:
Starting review

src/main/webapp/course.html
```diff
@@ -163,18 +186,41 @@ <h1 class="p-4 text-center">Add Your Review</h1>
...
+                required
               />
             </div>
+            <div class="valid-tooltip">Looks good!</div>
+            <div class="invalid-tooltip">
+              Please input a valid difficulty score.
+            </div>
+            <div class="col">
+              <p class="lead">What was your grade in the class?</p>
+              <select id="grade" name="grade" class="form-control mb-5">
```
LukeGMTravis, line 199: Suggestion: "I" for Incomplete is a grade I've seen before. Maybe not in college though...? Unsure

src/main/webapp/course.html
```diff
@@ -85,7 +85,12 @@ <h1 class="p-4 text-center">Add Your Review</h1>
                 class="form-control mb-5"
                 rows="5"
                 placeholder="Write Opinion Here"
+                required
               ></textarea>
+              <div class="valid-tooltip">Looks good!</div>
```
LukeGMTravis, line 90: I don't see anywhere selectively hiding/showing these tooltips, where is this handled?

src/main/webapp/course.html
lqueipom: Great idea will take care of that!

src/main/webapp/course.html
lqueipom: It was supposed to automatically take care of it adding the class="needs-validate" to the form and adding the bootstrap source to the HTML. I have encountered problem with getting this to display tho so I will work on this on a separate PR to get this to work.

PR #50, opened 2020-08-20
Fuzzyname

ninaprabhu:


src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -75,19 +75,47 @@ public JSONObject getMatchingCourses(HttpServletRequest request) {
...
         List<String> courseNums = new ArrayList<>();
-        courseNums.add(department + " " + String.valueOf(courseNum + 1));
         courseNums.add(name);
-        courseNums.add(department + " " + String.valueOf(courseNum - 1));
-        nameFilter = new FilterPredicate("name", FilterOperator.IN, courseNums);
+        try {
+          String department = splitName[0];
+          int courseNum = Integer.parseInt(splitName[1]);
+          if (school.equals(
+              "mit")) { // Account for the fact that some course numbers have leading zeroes (e.g.
```
LukeGMTravis, line 96: For long comments like this (and often for comments in general), we tend to put them on their own lines to avoid weird wrapping like this.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -75,19 +75,47 @@ public JSONObject getMatchingCourses(HttpServletRequest request) {
...
+            // 6.006)
+            courseNums
```
LukeGMTravis, line 98: Is a courseNum of zero possible? What about nine? Do these edge cases behave the way you want them to?

src/main/webapp/search-script.js
```diff
@@ -23,8 +23,7 @@ async function showCourses() {
   url.searchParams.set("profName", profName);
   url.searchParams.set("units", units);
   url.searchParams.set("term", termName);
-  url.searchParams.set("schoolName", school);
-
+  url.searchParams.set("school-name", school);
```
LukeGMTravis, line 26: Why is this hyphenated but "profName" above isn't?

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: Added in a case for 0; 9 behaves the way I want it to.

src/main/webapp/search-script.js
ninaprabhu: Changed this to be hyphenated throughout

PR #61, opened 2020-08-21
Adds Functionality to Professor Page

hailed748:
Professor Page now pulls data from ProfessorData servlet

### DIFF
--- a/src/main/webapp/course.html
+++ b/src/main/webapp/course.html
@@ -37,6 +37,9 @@
       <div class="container">
         <h1 class="display-4 text-center" id="course-name">Course Name</h1>
         <p class="lead text-center" id="term-name">Term Name</p>
+        <a id="prof-link"
+          ><p class="lead text-center" id="prof-name">Professor Name</p></a
+        >
         <p class="text-muted text-center" id="num-enrolled">Num Enrolled</p>
       </div>
     </div>
@@ -59,14 +62,6 @@
       <div class="my-3 p-3 bg-white" id="course-comments">
         <h1 class="border-bottom border-gray p-5 mb-0 text-center">
           Course Comments
-        </h1>
-      </div>
-    </div>
-
-    <div class="plr-20 container-fluid">
-      <div class="my-3 p-3 bg-white" id="prof-comments">
-        <h1 class="border-bottom border-gray pb-5 mb-0 text-center">
-          Professor Comments
         </h1>
       </div>
     </div>
@@ -176,6 +171,9 @@
             Submit
           </button>
         </form>
+        <button id="myBtn" type="button" class="btn btn-lg btn-outline-success">
+          Submit
+        </button>
       </div>
     </div>
     <script src="https://cdn.jsdelivr.net/npm/jstat@1.9.3/dist/jstat.min.js"></script>
--- a/src/main/webapp/course.html
+++ b/src/main/webapp/course.html
@@ -163,15 +163,12 @@
               />
             </div>
           </div>
-          <button
-            onclick="passRatingProperties()"
-            type="button"
-            class="btn btn-lg btn-outline-success"
-          >
-            Submit
-          </button>
         </form>
-        <button id="myBtn" type="button" class="btn btn-lg btn-outline-success">
+        <button
+          id="form-submit"
+          type="button"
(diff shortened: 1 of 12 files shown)


LukeGMTravis:
Starting review

src/main/webapp/course.html
```diff
@@ -176,6 +171,9 @@ <h1 class="p-4 text-center">Add Your Review</h1>
             Submit
           </button>
         </form>
+        <button id="myBtn" type="button" class="btn btn-lg btn-outline-success">
```
LukeGMTravis, line 174: "myBtn" strikes me as a non-descriptive id name

src/main/webapp/courseLoader.js
```diff
@@ -8,13 +8,23 @@
...
+    url.searchParams.set("prof-key", profKey);
+    url.searchParams.set("prof-name", profName);
+    profLink.setAttribute("href", url);
+  }
+
   function fillTitles() {
     fetch(`/term-info?term-key=${termKey}&course-key=${courseKey}`)
       .then((response) => response.json())
       .then((termInfo) => {
         document.getElementById("course-name").innerHTML = termInfo[0];
```
LukeGMTravis, line 23: Ideally these were all named properties rather than indexes. It would be much more clear (and harder to screw up) if this was instead `document.getElementById("term-name").innerHTML = termInfo.name;`.

src/main/webapp/courseLoader.js
```diff
@@ -475,6 +484,10 @@
     postRatingProperties(urlAndData[0], urlAndData[1]);
   }
 
+  document
+    .getElementById("myBtn")
+    .addEventListener("click", passRatingProperties);
```
LukeGMTravis, line 489: Probably cleaner to have an onclick attribute on the html

src/main/webapp/profLoader.js
```diff
@@ -0,0 +1,128 @@
...
+    const average = (list) =>
+      list.reduce((prev, curr) => prev + curr, 0) / list.length;
+
+    const courseList = [];
+    const difficultyAvgList = [];
+
+    for (let dataHolder of profData) {
+      courseList.push(dataHolder.term);
+      difficultyAvgList.push(
+        average(dummyGradeData.concat(dataHolder.difficultyList)) + 3
```
LukeGMTravis, line 37: What is this 3 for? Might be good to have a comment

src/main/webapp/courseLoader.js
```diff
@@ -194,7 +203,7 @@
 
   async function makeTermPerceptionChart(termDataObject) {
     const average = (list) =>
```
LukeGMTravis, line 205: Looks like this average function gets used in multiple functions. Probably best to only define it once instead of repeating yourself. Ideally this would be imported from a utilities file or something like that. For here it's fine to just define it once per file instead if you want.

src/main/webapp/profLoader.js
```diff
@@ -0,0 +1,128 @@
...
+      list.reduce((prev, curr) => prev + curr, 0) / list.length;
+
+    const courseList = [];
+    const perceptionAvgList = [];
+    const commentList = [];
+
+    for (let dataHolder of profData) {
+      courseList.push(dataHolder.term);
+      perceptionAvgList.push(
+        average(dummyGradeData.concat(dataHolder.perceptionList)) + 7
```
LukeGMTravis, line 72: Same question as above, why 7 here? Comment would help

src/main/webapp/courseLoader.js
hailed748: We had an onclick but for some reason it would not work so Laura decided to move to this implementation

src/main/webapp/profLoader.js
hailed748: will add, its just to make the dummy data look different from the other data

src/main/webapp/profLoader.js
hailed748: will add, its just to make the dummy data look different from the other data

src/main/webapp/courseLoader.js
hailed748: Will do in a separate PR that refactors some stuff

PR #64, opened 2020-08-24
Change form styling/parameters

ninaprabhu:


src/main/webapp/add-course.html
```diff
@@ -38,9 +38,10 @@
                     </select>
                 </div>
                 <textarea name="school-name" id="school-name" placeholder="School Name" maxlength=1500></textarea>
-                <input type="submit"/>
+                <input type="submit"/ id="add-button">
```
hailed748, line 41: super nit: / is in the wrong place

src/main/webapp/search-style.css
```diff
@@ -26,7 +26,7 @@ textarea, select, fieldset {
   width: 50%;
 }
 
-#search-button, #logout-button {
+#search-button, #logout-button, #private-logout-button, #add-to-search, #search-to-add, #add-button {
```
hailed748, line 29: This is a complete suggestion but you could just make this a class like "custom-button" and just give all of them the same class that way if you want to override one specifically you could do it through the ID

LukeGMTravis:
Starting review

src/main/webapp/add-course.html
LukeGMTravis: I'd say this is more significant than a nit actually, since it's not just stylistic but actually malformed. (The reason this doesn't actually break things is because web browsers have been designed to accept html with mistakes like this.)

src/main/webapp/search-script.js
```diff
@@ -63,7 +60,7 @@ async function addCourse() {
   const courseName = document.getElementById("course-name").value;
   const profName = document.getElementById("prof-name").value;
   const termName = document.getElementById("term").value;
-  const schoolName = document.getElementById("schoool-name").value;
```
LukeGMTravis, line 66: Lol nice catch

src/main/webapp/search-style.css
LukeGMTravis: This is a good suggestion, though I'd add that you should make the class something more descriptive than just "custom-button".

src/main/webapp/add-course.html
ninaprabhu: omg totally didn't notice that - thanks!

PR #65, opened 2020-08-25
Adds course links and modifies graph legend

hailed748:
Adds course links and modifies graph legend 

image

### DIFF
--- a/src/main/webapp/profLoader.js
+++ b/src/main/webapp/profLoader.js
@@ -50,7 +50,7 @@
     for (let dataHolder of profData) {
       courseList.push(`${dataHolder.course} - ${dataHolder.term}`);
       difficultyAvgList.push(
-        average(dummyGradeData.concat(dataHolder.difficultyList)) + 3
+        average(dummyGradeData.concat(dataHolder.difficultyList))
       );
     }
 
@@ -82,7 +82,7 @@
     for (let dataHolder of profData) {
       courseList.push(`${dataHolder.course} - ${dataHolder.term}`);
       perceptionAvgList.push(
-        average(dummyGradeData.concat(dataHolder.perceptionList)) + 7
+        average(dummyGradeData.concat(dataHolder.perceptionList))
       );
 
       for (let comment of dummyComments.concat(dataHolder.commentsList)) {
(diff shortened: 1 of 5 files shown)


hailed748:
Was already reviews by both Nina and Laura yesterday, but I had to delete the PR and repost it due to some commit errors that occurred.

Tetheta:
Starting Review

src/main/webapp/profLoader.js
```diff
@@ -11,14 +11,30 @@
   const profKey = urlParams.get("prof-key");
   const profName = urlParams.get("prof-name");
 
+  function fillTitle(profData) {
+    document.getElementById("prof-name").innerHTML = profName;
+    linkContainer = document.getElementById("link-container");
```
Tetheta, line 16: Is this a local variable, or a global you declare somewhere? If the latter I'm not seeing it.
I'd prefer to declare this here as a const and just re-fetch from the document if you need it elsewhere.

FYI JavaScript lets you declare variables with no var/etc in front, but they become vars by default.

src/main/webapp/profLoader.js
```diff
@@ -11,14 +11,30 @@
...
+
+    for (let dataHolder of profData) {
+      const newLink = document.createElement("a");
+      newLink.innerHTML = `${dataHolder.course} - ${dataHolder.term}`;
```
Tetheta, line 20: FYI this would be trivially easy to execute a XSS attack with. They could override your variable here in dev tools and insert their own script tag which would then run with full page & origin permissions.

Not a concern for this app, but good to know if you're ever making a production website.

src/main/webapp/profLoader.js
```diff
@@ -64,9 +80,9 @@
     const commentList = [];
 
     for (let dataHolder of profData) {
-      courseList.push(dataHolder.term);
+      courseList.push(`${dataHolder.course} - ${dataHolder.term}`);
       perceptionAvgList.push(
-        average(dummyGradeData.concat(dataHolder.perceptionList))
+        average(dummyGradeData.concat(dataHolder.perceptionList)) + 7
```
ninaprabhu, line 85: why 7? Just wondering

hailed748: Its just dummy data to make the graphs data look real lol

PR #68, opened 2020-08-25
course add/form changes

ninaprabhu:


Tetheta:
Starting review

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -223,24 +223,49 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    List<Filter> filters = new ArrayList<>();
+    filters.add(new FilterPredicate("search-name", FilterOperator.EQUAL, name.toLowerCase()));
+    filters.add(new FilterPredicate("search-professor", FilterOperator.EQUAL, prof.toLowerCase()));
+    filters.add(new FilterPredicate("units", FilterOperator.EQUAL, units));
+    filters.add(new FilterPredicate("term", FilterOperator.EQUAL, term));
+    filters.add(new FilterPredicate("school", FilterOperator.EQUAL, school));
+
+    Query courseQuery = new Query("Course-Info");
+    if (!filters.isEmpty()) {
+      if (filters.size() == 1) {
```
Tetheta, line 261: When would filters be empty or size of only 1? Aren't you adding 5 filters directly above here?

src/main/webapp/search-style.css
```diff
@@ -26,7 +26,7 @@ textarea, select, fieldset {
   width: 50%;
 }
 
-#search-button, #logout-button {
+#search-button, #logout-button, #private-logout-button, #add-to-search, #search-to-add, #add-button {
```
Tetheta, line 29: Might think about extracting this to a class name in the future if it further expands.

ninaprabhu: Made all of this into the class course-button

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: good point, I think that was just a reflex stuck from the search in doGet

PR #71, opened 2020-08-26
Adding Button To Fetch Latest Rating

lqueipom:
Created:
- LatestRating Servlet to get the Rating Entity of a user. 
- ratingScript JS file to separate my code from Dagmawi's.

Added:
- In ratingScript JS, I added getLatestRating() which interacts with my LatestRating Servlet and populates course.html with the Rating it finds or displays a message if it does not find anything. 
- Button to course.html to fetch rating. 
- Added a "translate" property to the Rating Entity to be able to autofill all fields of the form.


src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -158,6 +158,7 @@ public void addTermRating(HttpServletRequest request, DatastoreService datastore
     termRatingEntity.setProperty("comments-professor", professorFeedback);
     termRatingEntity.setProperty("score-professor", professorScore);
     termRatingEntity.setProperty("perception-professor", professorRating);
+    termRatingEntity.setProperty("translation", Boolean.toString(translate));
```
ninaprabhu, line 161: What if someone types something in a different language but then doesn't check the translate button? Is there a way to detect if the comment is in a different language and automatically translate then?

src/main/java/com/google/sps/servlets/LatestRating.java
```diff
@@ -0,0 +1,53 @@
...
+      Entity previousRatingEntity = latestRatingQuery.get(0);
+      previousRatingInfo.put("termInput", previousRatingEntity.getProperty("comments-term"));
+      previousRatingInfo.put("profInput", previousRatingEntity.getProperty("comments-professor"));
+      previousRatingInfo.put("difficulty", previousRatingEntity.getProperty("difficulty"));
+      previousRatingInfo.put("hours", previousRatingEntity.getProperty("hours"));
+      previousRatingInfo.put("ratingTerm", previousRatingEntity.getProperty("perception-term"));
+      previousRatingInfo.put(
+          "ratingProf", previousRatingEntity.getProperty("perception-professor"));
+      previousRatingInfo.put("grade", previousRatingEntity.getProperty("grade"));
+      previousRatingInfo.put("translation", previousRatingEntity.getProperty("translation"));
```
ninaprabhu, line 42: is there a reason you don't name the properties the same thing and set the entities equal to each other? or is there a reason that wouldn't work

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: So, originally I was doing this to save on operation time and maybe if you automatically translated something from English to English it could change in some capacity if there was like slang or something so wanted to let them choose. Let's check with Luke and Travis on this to see what they thinkk :)

src/main/java/com/google/sps/servlets/LatestRating.java
lqueipom: So in this case, I was trying to make the names similar to how they are inputted in the actual html file (which is with term input, prof input, etc). So I was trying to map from this one to the JS one.

lqueipom: But to be honest, it would be easier if they could all be changed to match everywhere so will check with Dagmawi in case he uses this Entity somewhere.

lqueipom: Talked with Dagmawi and we decided that for now we would stick with how they are formatted for now to avoid confusion

src/main/java/com/google/sps/servlets/DataServlet.java
lqueipom: Checked! I think its better if we keep this as it is for the moment (talked with Luke and Travis and right now is not a priority)

LukeGMTravis:
Starting review

src/main/webapp/course.html
```diff
@@ -212,6 +212,14 @@ <h1 class="p-4 text-center">Add Your Review</h1>
         >
           Submit
         </button>
+        <div id="message-rating"></div>
+        <button
+          id="latest-rating"
+          type="button"
+          class="btn btn-lg btn-outline-success"
+        >
+          Autofill Form With Previous Rating.
```
LukeGMTravis, line 221: Nit: It's not technically *auto*filling if you have to push a button to do it right? : P

src/main/webapp/course.html
```diff
@@ -212,6 +212,14 @@ <h1 class="p-4 text-center">Add Your Review</h1>
         >
           Submit
         </button>
+        <div id="message-rating"></div>
```
LukeGMTravis, line 215: I don't think "message-rating" isn't a very clear id name in this case

src/main/webapp/courseLoader.js
```diff
@@ -380,77 +380,4 @@
       createComment(comment, isCourse);
     }
   }
-
```
LukeGMTravis, line 383: In the future, when simultaneously moving a lot of code like this to a new file and then adding to it, please do this in 2 PRs. That would make it much easier for your reviewers to see exactly what additions you made and what was simply moving code between files.

lqueipom: Note taken, sorry about that! :)

src/main/webapp/course.html
lqueipom: Hahaha, you are right! Changing name now!

PR #76, opened 2020-08-27
Improving Form Styling

lqueipom:
Added:
   -To course.html added the pop up messages for the optional fields (grade + translation). 
  - To course-style.css added a class to move the pop ups to right underneath the form fields. 
  - To ratingScript added to getLatestRating() code that added and removed classes from the message-rating div to change its color. The color was dependent on which step of rating retrieval the program was on. 
 
When merging my LatestRatingServlet file got committed again, feel free to ignore it if has been reviewed in my other PR. Sorry about that!

Screenshot 2020-08-27 at 12 36 26 PM


Tetheta:
Starting review

src/main/java/com/google/sps/servlets/LatestRating.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    JSONObject previousRatingInfo = new JSONObject();
+    Key parentTermKey = KeyFactory.stringToKey((String) request.getParameter("term-key"));
+    String userId = (String) request.getParameter("reviewer-id");
+    List<Entity> latestRatingQuery = getLatestRating(userId, parentTermKey);
+
+    if (!latestRatingQuery.isEmpty()) {
+      Entity previousRatingEntity = latestRatingQuery.get(0);
+      previousRatingInfo.put("termInput", previousRatingEntity.getProperty("comments-term"));
```
Tetheta, line 34: Do all these work fine if the previous rating entity only has a few fields populated?

src/main/webapp/courseLoader.js
```diff
@@ -380,77 +380,4 @@
       createComment(comment, isCourse);
     }
   }
-
```
Tetheta, line 383: This is all content we do want to delete right?

src/main/webapp/ratingScript.js
```diff
@@ -0,0 +1,128 @@
...
+    document.getElementById("rating-prof").value = formInfo["ratingProf"];
+    document.getElementById("hours").value = formInfo["hours"];
+    document.getElementById("difficulty").value = formInfo["difficulty"];
+    document.getElementById("grade").value = formInfo["grade"];
+    document.getElementById("translate").value = formInfo["translation"];
+
+    document.getElementById("message-rating").innerHTML =
+      "Your form has been populated!";
+    document
+      .getElementById("message-rating")
```
Tetheta, line 114: FYI everytime you do this you force the DOM to layout itself, then when you modify the class list it has to reflow. 
This is called layout thrashing.

Idelaly, declare a variable up top with messageRatingElement inside, then use that everywhere. It's also a bit cleaner code.

src/main/webapp/ratingScript.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+document
+  .getElementById("form-submit")
+  .addEventListener("click", passRatingProperties);
+
+document
+  .getElementById("latest-rating")
+  .addEventListener("click", getLatestRating);
```
Tetheta, line 128: These are happening globally? You could just add this to the html directly, onClick="passRatingProperties()" etc I think. 
Though if this works I guess it's fine.

src/main/webapp/courseLoader.js
lqueipom: Yes! It is still being used, I just wanted to separate mine and Dagmawi's stuff in different files.

src/main/webapp/ratingScript.js
lqueipom: Ohh did not know that, changing that now!

src/main/webapp/ratingScript.js
lqueipom: For some reason that was not working for us before, but I will try to add it back and see if that got fixed!

src/main/java/com/google/sps/servlets/LatestRating.java
lqueipom: I just realized that it does not, I am going to make some changes so that grades is the only optional field and that can proceed normally.

PR #79, opened 2020-08-27
Removes Dummy Data & Handles No Data Case

hailed748:
Modifes: LiveCourseData.java, LiveCourseDataTest.java, course.html. courseLoad.js, profLoader.js

1) Removed all dummy data from the professor page and course page 
2) Modifies histograms so they do not display unless there is 2 or more data points because graph breaks if any less
3) Adds a tooltip on the grade distribution nothing that if all grades are the same no distribution will show because you need at least two distinct points
4) Modifies properties that were mixed up on LiveCourseData


Tetheta:
Starting review

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -36,14 +36,16 @@ public TermDataHolder getAllDataFromTerm(DatastoreService db, HttpServletRequest
 
     termDataHolder.setHoursList(getDataFromTermRating(db, termKey, "hours"));
     termDataHolder.setDifficultyList(getDataFromTermRating(db, termKey, "difficulty"));
-    termDataHolder.setTermScoreList(getDataFromTermRating(db, termKey, "score-term"));
-    termDataHolder.setTermPerceptionList(getDataFromTermRating(db, termKey, "perception-term"));
+    termDataHolder.setTermScoreList(getDataFromTermRating(db, termKey, "perception-term"));
+    termDataHolder.setTermPerceptionList(getDataFromTermRating(db, termKey, "score-term"));
```
Tetheta, line 40: Why are we swapping terms here but not what we're setting?

src/main/java/com/google/sps/servlets/LiveCourseData.java
```diff
@@ -36,14 +36,16 @@ public TermDataHolder getAllDataFromTerm(DatastoreService db, HttpServletRequest
...
     termDataHolder.setProfessorPerceptionList(
+        getDataFromTermRating(db, termKey, "score-professor"));
+    termDataHolder.setProfessorScoreList(
         getDataFromTermRating(db, termKey, "perception-professor"));
```
Tetheta, line 44: Same question here

src/main/webapp/courseLoader.js
```diff
@@ -35,6 +35,7 @@
     fetch(`/term-data?term-key=${termKey}`)
       .then((response) => response.json())
       .then((data) => {
+        console.log(data);
```
Tetheta, line 38: Remove please

src/main/webapp/courseLoader.js
```diff
@@ -52,121 +53,124 @@
...
-        fontSize: 16,
-        bold: false,
-        italic: false,
-      },
-      legend: { position: "none" },
-      hAxis: {
-        title: "Hours",
+    if (termDataObject.hoursList.length < 2) {
+      const hourChartHolder = document.getElementById("hours-chart");
+      hourChartHolder.remove();
```
Tetheta, line 62: You could make this a one-liner if you want, no need for a variable to be declared.

src/main/webapp/courseLoader.js
```diff
@@ -259,13 +261,39 @@
...
+      "C-": 72.5,
+      "D+": 67.5,
+      "D": 65,
+      "D-": 62.5,
+      "F": 55,
+    };
+
+    const numericalGrade = [];
+    for (let grade of termDataObject.gradesList) {
+      if (grade != "I") {
```
Tetheta, line 284: Are we guaranteed gradesList will just be either things in gradeMapper or "I"?
If there's any weird data here it will still get pushed.

hailed748: Yes we are guaranteed this because the grade is selected via drop-down so the user is limited in what they input

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: This is due to a misunderstanding I had with Laura on what each thing meant. When she creates the rating entity, the score-term property on the rating entity actually refers to the perception of a term because the API calls it a "perception score"  and the "perception-term" property relates the the term-score. It's kinda backwards so I mixed them up, but because we already have some seeded data I went for this quick fix rather than having her rename the entity property

Tetheta: Hmm. Well it's definitely super confusing to look at. Is there a TODO somewhere to fix this? It'd be good to get this finished before reviewers look at the completed code/etc.

src/main/webapp/courseLoader.js
Tetheta: Doesn't appear to have been removed?

hailed748:
Oh my bad committed the changes forgot to push them

src/main/java/com/google/sps/servlets/LiveCourseData.java
hailed748: I will speak with Laura first thing tomorrow and get it fixed

PR #84, opened 2020-08-31
Adding More Tests for DataServlet

lqueipom:
While deploying I realized that the current DataServletTest was not testing all possible scenarios. Hence, I added three more tests testing translation, adding different term ratings for the same course, and adding term ratings to different courses but with the same term name. 

Modified: 
- Added this.translation to the constructor of DataServlet to be able to pass the mock TranslationService from DataServletTest. 
- Made the corresponding changes to getTranslatedText() in DataServlet.
- Fixed a small naming bug in course.html and ratingScript.js


LukeGMTravis:
Starting review

src/main/webapp/course.html
```diff
@@ -217,13 +217,13 @@ <h1 class="p-4 text-center">Add Your Review</h1>
             <div class="col">
               <p class="lead">Want to Translate to English?</p>
               <select class="custom-select mb-5" id="translate">
-                <option value="false" disabled selected>Wants to translate?</option>
+                <option value="false" disabled selected>
+                  Wants to translate?
```
LukeGMTravis, line 221: "Want"

src/test/java/com/google/sps/servlets/DataServletTest.java
```diff
@@ -171,6 +178,235 @@ public void addTermRating_overwritingExistingTermRating() throws IOException {
...
+    HttpServletRequest request = Mockito.mock(HttpServletRequest.class);
+    Reader reader = new FileReader("src/main/webapp/WEB-INF/testTranslation.txt");
+    BufferedReader bufferedReader = new BufferedReader(reader);
+    when(request.getReader()).thenReturn(bufferedReader);
+
+    AnalyzeSentimentResponse response =
+        AnalyzeSentimentResponse.newBuilder()
+            .setDocumentSentiment(
+                Sentiment.newBuilder()
+                    .setScore((float) -0.8999999761581421)) // Sentiment Score of Text.
```
LukeGMTravis, line 194: This float is pretty long and arbitrary. Might want to have it as a const rather than typing it out like this each time it's used.

lqueipom: Just put it in a variable, for some reason when I try to change it to a double from a float it shortens the number itself. Hence, when I am doing assert I am typing it out in the asserts.

PR #86, opened 2020-08-31
Putting translation as a checkbox

lqueipom:
Modified:
- Now translation will be done if you check the "translate" checkbox in course.html 
- ratingProperties in the ratingScript.js as now we will determine if we will translate with the .checked attribute


LukeGMTravis:
Starting review

PR #88, opened 2020-08-31
change various pages' visibility/functionality based on allowlist

ninaprabhu:
1. Changed format of whitelist object - combined ability to check for verified user and white-list verified user in one pass through the servlet, instead of having to make two calls. 

2. Made reports-list.html admin-restricted.

3. Hid "add course" button from non-whitelisted users.

### DIFF
--- a/src/main/java/com/google/sps/servlets/LoginServlet.java
+++ b/src/main/java/com/google/sps/servlets/LoginServlet.java
@@ -5,7 +5,6 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
 import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken.Payload;
 import com.google.api.client.googleapis.auth.oauth2.GoogleIdTokenVerifier;
 import com.google.api.client.json.gson.GsonFactory;
-import com.google.sps.data.LoginObject;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


Tetheta:
Starting review

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -28,39 +27,51 @@
...
       GoogleIdToken idToken = verifier.verify(idTokenString);
       if (idToken != null) {
         Payload payload = idToken.getPayload();
         String userId = payload.getSubject();
-        if (checkWhitelist && !(whitelist.contains(payload.getEmail()))) {
-          return new LoginObject("", false, "NOT AUTHORIZED");
+        response.put("userId", userId);
+        response.put("verified", "true");
+        if (whitelist.contains(payload.getEmail())) {
+          response.put("whitelist", true);
```
Tetheta, line 49: FYI as a community we're trying to remove references to white/blacklist and instead use allowlist and blocklist. 
Might be good to add at least a TODO to that note. Ideally something to refactor before we send out code (it's a Google mandate now I think that all new code should be properly renamed)

ninaprabhu: Oh thanks for pointing that out! Will do right now

PR #89, opened 2020-08-31
Implement substring search

ninaprabhu:
Implemented fuzzy search through regexes.

FILES CHANGED: SearchServlet.java.

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,14 +143,6 @@
             <trimStackTrace>false</trimStackTrace>
         </configuration>
       </plugin>
-      <!-- <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.2.0</version>
-        <configuration>
-            <trimStackTrace>false</trimStackTrace>
-        </configuration>
-    </plugin> -->
     <plugin>
       <groupId>com.cosium.code</groupId>
       <artifactId>maven-git-code-format</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -139,7 +139,18 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.22.1</version>
+        <configuration>
+            <trimStackTrace>false</trimStackTrace>
+        </configuration>
       </plugin>
+      <!-- <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.2.0</version>
+        <configuration>
+            <trimStackTrace>false</trimStackTrace>
+        </configuration>
+    </plugin> -->
     <plugin>
       <groupId>com.cosium.code</groupId>
       <artifactId>maven-git-code-format</artifactId>
--- a/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -196,6 +196,7 @@ public class SearchServlet extends HttpServlet {
     Boolean hasProf = !request.getParameter("prof-name").isEmpty();
     Pattern namePattern = Pattern.compile("/(?=a)b/"); // Always false.
     Pattern profPattern = Pattern.compile("/(?=a)b/"); // Always false.
+
     if (hasName) {
       String name = request.getParameter("course-name").toLowerCase();
       namePattern = Pattern.compile("^.*" + name + ".*$");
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -139,7 +139,18 @@
...
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.22.1</version>
+        <configuration>
+            <trimStackTrace>false</trimStackTrace>
+        </configuration>
       </plugin>
+      <!-- <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.2.0</version>
```
lqueipom, line 149: Comment out/Delete these lines

LukeGMTravis:
Starting review

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -170,9 +184,50 @@ public JSONObject getMatchingCourses(HttpServletRequest request) {
...
+
     return results;
   }
 
+  /* Does substring matching through regex. */
+  private List<Entity> processResults(HttpServletRequest request, List<Entity> fuzzyResults) {
+    List<Entity> processed = new ArrayList<>();
+    Boolean hasName = !request.getParameter("course-name").isEmpty();
+    Boolean hasProf = !request.getParameter("prof-name").isEmpty();
+    Pattern namePattern = Pattern.compile("/(?=a)b/"); // Always false.
```
LukeGMTravis, line 196: Since these are only used if they get reassigned you can simply initialize here (`Pattern namePattern;`) without assigning.

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -170,9 +184,50 @@ public JSONObject getMatchingCourses(HttpServletRequest request) {
...
+      } else {
+        nameMatches = true;
+      }
+      if (hasProf) {
+        Matcher profMatcher = profPattern.matcher(searchProf);
+        profMatches = profMatcher.matches();
+      } else {
+        profMatches = true;
+      }
+      if (nameMatches && profMatches && (hasName || hasProf)) {
```
LukeGMTravis, line 224: Do you need to check if the request had a name or prof?

src/main/webapp/login-script.js
```diff
@@ -48,6 +51,25 @@ async function signInPrivate(googleUser) {
   }
 }
 
+async function signInReports(googleUser) {
```
LukeGMTravis, line 54: Nit: I feel this would be a bit clearer if it were "reportsSignIn" perhaps

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: if it didn't have either, then nameMatches && profMatches would be true, but I don't want to add it to the list. So I want to make sure that both of them aren't true just because the parameter didn't exist.

src/main/java/com/google/sps/servlets/SearchServlet.java
ninaprabhu: it gives me an error as "namePattern and profPattern may not have been initialized"

src/main/webapp/login-script.js
ninaprabhu: Makes sense, changed signInPrivate to fit this convention too

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -155,8 +169,9 @@ public JSONObject getMatchingCourses(HttpServletRequest request) {
         filters.add(unitsFilter);
       }
     }
-
-    return filters;
+    // filterLists.add(filters);
```
LukeGMTravis, line 172: Should these be removed?

src/main/java/com/google/sps/servlets/SearchServlet.java
LukeGMTravis: Ooh right, cause you assign in an if statement... Hm... I suppose this is alright then.

src/main/java/com/google/sps/servlets/SearchServlet.java
LukeGMTravis: Why don't you want to add to the list if the search query had no name or prof?

ninaprabhu:
@LukeGMTravis I can't reply to your comment directly but:
"Why don't you want to add to the list if the search query had no name or prof?"
This function deals with the fuzzyFilters list of the filterPair object generated earlier in the program. The particular course in question would have already been added in the mainFilters attribute of that object, so we do not need to add it twice. I'm just doing a check here to avoid needlessly increasing the size of the list returned in this function.

PR #93, opened 2020-09-01
Adding submission checker

lqueipom:
Added:
- Added submission alert to course.html to notify whether submission was valid or invalid.
- Function validSubmission() to ratingScript.js which will be called before each attempt at submission.
    - If valid, it will proceed with submission as normally.
    - If not, it will prevent submission and ask the user to check the form fields. 

Modified:
- Removed label from checkbox in course.html as it is not necessary for this case.

### DIFF
--- a/src/main/webapp/ratingScript.js
+++ b/src/main/webapp/ratingScript.js
@@ -130,7 +130,6 @@ function validSubmission() {
   for (i = 0; i < allFormFields.length; i++) {
     // If any field is not field out properly.
     if (!allFormFields[i].checkValidity()) {
-      console.log(allFormFields[i]);
       successfulSubmissionMessage.innerHTML =
         "Invalid Submission, check form fields";
       return false;
--- a/src/main/webapp/ratingScript.js
+++ b/src/main/webapp/ratingScript.js
@@ -59,7 +59,7 @@ function addingAttributesToURL(ratingProperties) {
 }
 
 async function passRatingProperties() {
-  if (validSubmission() == true) {
+  if (validSubmission()) {
     document.getElementById("retrieve-last-rating-message").innerHTML = "";
     const urlAndData = await getRatingPropertiesToStore();
     postRatingProperties(urlAndData[0], urlAndData[1]);
--- a/src/main/webapp/ratingScript.js
+++ b/src/main/webapp/ratingScript.js
@@ -59,10 +59,13 @@ function addingAttributesToURL(ratingProperties) {
 }
 
 async function passRatingProperties() {
-  if (validSubmission()) {
+  if (isValidSubmission()) {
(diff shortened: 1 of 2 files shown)


src/main/webapp/ratingScript.js
```diff
@@ -59,9 +59,11 @@ function addingAttributesToURL(ratingProperties) {
 }
 
 async function passRatingProperties() {
-  document.getElementById("retrieve-last-rating-message").innerHTML = "";
-  const urlAndData = await getRatingPropertiesToStore();
-  postRatingProperties(urlAndData[0], urlAndData[1]);
+  if (validSubmission() == true) {
```
ninaprabhu, line 62: if (validSubmission())

lqueipom: Thanks!!

LukeGMTravis:
Starting review

src/main/webapp/ratingScript.js
```diff
@@ -110,6 +112,34 @@ async function getLatestRating() {
...
+    document.getElementById("term-input"),
+    document.getElementById("prof-input"),
+    document.getElementById("rating-term"),
+    document.getElementById("rating-prof"),
+    document.getElementById("hours"),
+    document.getElementById("difficulty"),
+    document.getElementById("grade"),
+  ];
+
+  let i;
```
LukeGMTravis, line 126: You can declare this in the for loop

src/main/webapp/ratingScript.js
```diff
@@ -110,6 +112,34 @@ async function getLatestRating() {
...
+  let successfulSubmissionMessage = document.getElementById(
+    "sucessful-submission-message"
+  );
+  for (i = 0; i < allFormFields.length; i++) {
+    // If any field is not field out properly.
+    if (!allFormFields[i].checkValidity()) {
+      successfulSubmissionMessage.innerHTML =
```
LukeGMTravis, line 133: Maybe "successfulSubmissionMessage" doesn't make as much sense here. Potential rename idea: "submissionStatusMessage"

src/main/webapp/ratingScript.js
```diff
@@ -110,6 +112,34 @@ async function getLatestRating() {
   }
 }
 
+function validSubmission() {
```
LukeGMTravis, line 115: Nit: Usually functions returning a boolean like this have names like "isValidSubmission"

src/main/webapp/ratingScript.js
LukeGMTravis, line 133: It's a little odd that a function returning a boolean indicating whether or not form fields are valid also has a side effect writing a message to the dom. Might want to pull out the side effect.

lqueipom: Makes sense, thanks for the comment making all the changes now!

PR #94, opened 2020-09-01
Resolves comparison chart bug

hailed748:
Modifies: courseLoader.js, profLoader.js

Resolved bug where comparison charts either do not show or have random numbers in the thousands. Turns our my average function was not flattening the list quick fix


LukeGMTravis:
Starting review

PR #95, opened 2020-09-01
Fixing Bugs Last Run

lqueipom:
Modified:
- In verify() in ratingScript.js the user ID was saved differently than before, so we changed it accordingly. Now it is, userInfo.userID.
- I noticed a problem with the close button and the change of colors for the alerts in course.html so for now I am removing it until I am 100% it will not cause our code to break.
- Noticed a slight syntax problem in DataServletTest so I erased it. 
- Modified how we get translate boolean in DataServlet.
- Made some changes to isValidSubmission() in ratingScript.js that were made in another branch.

### DIFF
--- a/src/main/webapp/course.html
+++ b/src/main/webapp/course.html
@@ -14,7 +14,7 @@
   </head>
   <body>
     <nav class="navbar navbar-expand-lg navbar-dark bg-dark">
-      <span class="navbar-brand mb-0 h1">App Name</span>
+      <span class="navbar-brand mb-0 h1">CourseFYI</span>
       <button
         class="navbar-toggler"
         type="button"
@@ -262,18 +262,10 @@
 
         <div
           id="retrieve-last-rating-container"
-          class="alert alert-light alert-dismissible fade show"
+          class="alert alert-info alert-dismissible fade show"
           role="alert"
         >
           <strong id="retrieve-last-rating-message"></strong>
-          <button
-            type="button"
-            class="close"
-            data-dismiss="alert"
-            aria-label="Close"
-          >
-            <span aria-hidden="true">&times;</span>
-          </button>
         </div>
 
         <div
--- a/src/main/webapp/course.html
+++ b/src/main/webapp/course.html
@@ -274,11 +274,11 @@
         </div>
 
         <div
-          id="successful-submission-container"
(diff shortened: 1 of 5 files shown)


src/main/webapp/ratingScript.js
```diff
@@ -103,13 +102,32 @@ async function getLatestRating() {
...
+    document.getElementById("rating-prof"),
+    document.getElementById("hours"),
+    document.getElementById("difficulty"),
+    document.getElementById("grade"),
+  ];
+
+  );
+  for (let i = 0; i < allFormFields.length; i++) {
+    // If any field is not field out properly.
+    if (!allFormFields[i].checkValidity()) {
```
ninaprabhu, line 123: for (let field of allFormFields)

Tetheta:
Starting review

src/main/webapp/course.html
```diff
@@ -269,18 +267,18 @@ <h1 class="p-4 text-center">Add Your Review</h1>
...
-            <span aria-hidden="true">&times;</span>
-          </button>
+        </div>
+
+        <div
+          id="submission-status-container"
+          class="alert alert-warning alert-dismissible fade show"
+          role="alert"
+        >
+          <strong id="submission-status-message"></strong>
```
Tetheta, line 281: FYI typically we bold with css, not with strong tags

src/main/webapp/ratingScript.js
```diff
@@ -103,13 +102,32 @@ async function getLatestRating() {
...
+  ];
+
+  );
+  for (let field of allFormFields) {
+    // If any field is not field out properly.
+    if (!field.checkValidity()) {
+      return false;
+    }
+    // If no field was submitted the wrong way.
+    return true;
```
Tetheta, line 127: Shouldn't this be outside the for loop? Right now as soon as there's a valid field it's going to return true for the whole function.

lqueipom: Ohh that's weird! I think this got modified by accident in my previous PR changing that now! :)

src/main/webapp/course.html
lqueipom: Good to know!! Just changed that as well!

